PR #6, opened 2020-09-04
Add support for password confirmation middleware.

jtolj:
This PR adds controller methods to support the 'password.confirm' middleware from Illuminate/Auth. 

The idea here is that this could eventually be used to implement a solution for issues like:
https://github.com/laravel/jetstream/issues/17
https://github.com/laravel/jetstream/pull/20

Where certain high risk actions (deleting an API key, deleting the user account, disabling 2FA, etc) could require a password verification prior to accessing the route. This is the pattern that Github follows (a re-auth redirect), and I think it's the most flexible approach.

Thanks for considering!

### DIFF
--- a/src/Http/Controllers/VerifyPasswordController.php
+++ b/src/Http/Controllers/VerifyPasswordController.php
@@ -2,17 +2,17 @@
 
 namespace Laravel\Fortify\Http\Controllers;
 
-use Illuminate\Http\Request;
-use Illuminate\Routing\Controller;
 use Illuminate\Contracts\Auth\StatefulGuard;
 use Illuminate\Contracts\Support\Responsable;
+use Illuminate\Http\Request;
+use Illuminate\Routing\Controller;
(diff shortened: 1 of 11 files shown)


src/Http/Controllers/VerifyPasswordController.php
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Verify the user's password.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Contracts\Support\Responsable
+     */
+    public function store(Request $request): Responsable
+    {
+        $username = config('fortify.username');
+        if ($status = $this->guard->validate([
```
m1guelpf, line 53:
```suggestion
        $username = config('fortify.username');

        if ($status = $this->guard->validate([
```

taylorotwell:
Well done. This saved me some time today.

PR #12, opened 2020-09-05
Remove serialize when encrypting secret and recovery codes

MrMicky-FR:
This remove the serialization when encrypting the secret and the recovery codes as it seems unnecessary because the secret and the recovery codes are already strings (the recovery codes are json encoded before encryption).

This also helps when interacting with the database from other languages as the serialize format is only used in PHP so it’s hard to parse in non-PHP applications.

And I think now is the only time it’s possible to change this because after Fortify is released it would require to decrypt and encrypt all users with 2FA in the database to make this simple change and it's almost impossible with lot of users.


MrMicky-FR:
Hi,
Would it be possible to know why it was closed ?
Thanks

PR #27, opened 2020-09-11
Allows email address column name to be changed.

adear11:
I am working to build a new application that uses an existing database that uses 'emailAddress' as the column name in the users table. 

This pull request adds a config value to allow the email address property to be customized instead of relying a hard coded property name. It follows the same logic that is used to allow the username field to be set in the config file.

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -6,7 +6,7 @@ return [
     'guard' => 'web',
     'passwords' => 'users',
     'username' => 'email',
-    'emailaddress' => 'email',
+    'email_address' => 'email',
     'home' => '/home',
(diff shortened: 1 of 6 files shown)


taylorotwell:
None of these changes have anything to do with column names. Just request variables.

adear11:
So is the only issue that I'm using the term 'column name' to reference the request variables?

That's a function of the specific problem I'm running into is that the request variable names (and thus the database query that is ultimately generated) don't match the my particular application.

I can certainly correct those references if that is what it takes to get this change merged. Or is there another solution to this problem that I'm overlooking?

PR #28, opened 2020-09-11
[1.x] Allow the expected email address request variable to be changed

adear11:
This pull requests gives the ability to change the expected name of the variable in the request that holds the email address when making requests for forgotten passwords and when setting a new password. As it is implemented currently, the request variable name `email` is hard coded into the controllers for both requesting a password reset link and for setting the new password after visiting the reset link. This pull request adds a new config option, similar to the `username` option that is currently there, to allow for changing the expected variable name.

The implementation maintains the `email` name as a default value so that any existing applications should not experience any BC breaks.

The ability to set the expected name of the email address request variable via a config value makes it much easier to integrate a new Laravel application that uses Fortify into an existing system that already has an established database schema that cannot be changed. Without the ability to change that hard coded value via config value, it appears as though the only solution to this problem would be to re-implement all of the password reset logic and controllers simply to change the one request variable.

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -6,6 +6,7 @@ return [
     'guard' => 'web',
     'passwords' => 'users',
     'username' => 'email',
+    'email_address' => 'email',
     'home' => '/home',
     'limiters' => [
         'login' => null,
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -6,7 +6,7 @@ return [
     'guard' => 'web',
     'passwords' => 'users',
     'username' => 'email',
-    'email_address' => 'email',
+    'email' => 'email',
     'home' => '/home',
     'limiters' => [
         'login' => null,
--- a/src/Fortify.php
+++ b/src/Fortify.php
@@ -46,6 +46,14 @@ class Fortify
     public static $registersRoutes = true;
 
     /**
+     * Get the name of the email address request variable.
+     */
+    public static function emailAddress()
+    {
+        return config('fortify.email_address', 'email');
+    }
+
+    /**
      * Get the username used for authentication.
      */
     public static function username(): string
--- a/src/Fortify.php
+++ b/src/Fortify.php
@@ -48,9 +48,9 @@ class Fortify
     /**
      * Get the name of the email address request variable.
      */
-    public static function emailAddress()
+    public static function email()
     {
-        return config('fortify.email_address', 'email');
+        return config('fortify.email', 'email');
     }
 
     /**
--- a/src/Http/Controllers/PasswordResetLinkController.php
+++ b/src/Http/Controllers/PasswordResetLinkController.php
@@ -33,13 +33,13 @@ class PasswordResetLinkController extends Controller
      */
     public function store(Request $request): Responsable
     {
-        $request->validate([Fortify::emailAddress() => 'required|email']);
+        $request->validate([Fortify::email() => 'required|email']);
 
         // We will send the password reset link to this user. Once we have attempted
         // to send the link, we will examine the response then see the message we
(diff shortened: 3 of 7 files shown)


config/fortify.php
```diff
@@ -6,6 +6,7 @@
     'guard' => 'web',
     'passwords' => 'users',
     'username' => 'email',
+    'email_address' => 'email',
```
driesvints, line 9: Can we name this `email` to be consistent with other apps/packages in the eco system?

driesvints (changes requested):

src/Fortify.php
```diff
@@ -45,6 +45,14 @@ class Fortify
      */
     public static $registersRoutes = true;
 
+    /**
+     * Get the name of the email address request variable.
+     */
+    public static function emailAddress()
```
driesvints, line 51:
```suggestion
    public static function email()
```

config/fortify.php
adear11: I have made the requested changes.

src/Http/Controllers/PasswordResetLinkController.php
```diff
@@ -32,13 +33,13 @@ public function create(Request $request): RequestPasswordResetLinkViewResponse
...
     {
-        $request->validate(['email' => 'required|email']);
+        $request->validate([Fortify::email() => 'required|email']);
 
         // We will send the password reset link to this user. Once we have attempted
         // to send the link, we will examine the response then see the message we
         // need to show to the user. Finally, we'll send out a proper response.
         $status = $this->broker()->sendResetLink(
-            $request->only('email')
+            $request->only(Fortify::email())
```
taylorotwell, line 42: Isn't this the only line that needs to change in this entire PR? Your database doesn't care what the POST variable is. Only what the column in the database is.

driesvints:
@adear11 Taylor will review your PR once he's ready.

adear11:
I would have expected that what the POST variable was named would not
matter. However, that does not appear to be the case.

Further down the call stack, the retrieveByCredentials method of
Illuminate\Auth\EloquentUserProvider
uses the key names to build the query to retrieve the user by email
address. This ultimately means that the name of the email address POST
variable has to match the database column.

The same thing is true for src/Http/Controllers/NewPasswordController since
it uses the same retrieveByCredentials method call.

From src/Illuminate/Auth/EloquentUserProvider.php
<https://github.com/laravel/framework/blob/18940d55a0f183a00a678628a35bb923bc06fc18/src/Illuminate/Auth/EloquentUserProvider.php#L106>

    public function retrieveByCredentials(array $credentials)
    {

        ...

        // First we will add each credential element to the query as a
where clause.
        // Then we can execute the query and, if we found a user, return it
in a
        // Eloquent User "model" that will be utilized by the Guard
instances.
        $query = $this->newModelQuery();

        foreach ($credentials as $key => $value) {
            if (Str::contains($key, 'password')) {
                continue;
            }

            if (is_array($value) || $value instanceof Arrayable) {
                $query->whereIn($key, $value);
            } else {
                $query->where($key, $value);
            }
        }

        return $query->first();
      }

adear11:
@taylorotwell 

I got an email that you commented on this PR, but I can't find your comment. So I'm replying here rather than directly to your comment.

I would have expected that what the POST variable was named would not matter. However, that does not appear to be the case.

Further down the call stack, the `retrieveByCredentials` method of `Illuminate\Auth\EloquentUserProvider` uses the key names to build the query to retrieve the user by email address. This ultimately means that the name of the email address POST variable has to match the database column.

The same thing is true for `src/Http/Controllers/NewPasswordController` since it uses the same `retrieveByCredentials` method call.

PR #13, opened 2020-09-05
Provide a way of customizing/disabling the default routes

MarkIvanowich:
A mirror of Jetstream's https://github.com/laravel/jetstream/pull/67

Enables a user to call Fortify::ignoreRoutes() in a provider's register method to disable the default routes and provide their own.
Also enables publishing of default routes into userland.

Relevant are https://github.com/laravel/fortify/issues/9, https://github.com/laravel/fortify/pull/10

PR #41, opened 2020-09-16
Make routes more dynamic

jorenn92:
Addition for issue #13, This makes customizing routes easier.
Jetstream routes are also defined this way.

For instance: I needed to customize the url paths. So i turned the default routes off with the Fortify::ignoreRoutes(); option in my provider and copied the routes file. After that, the controller paths were broken. This commit fixes that scenario.


jorenn92:
I'll need to check why the email verification routes don't work.

PR #43, opened 2020-09-17
[1.x] tfa should not be allowed to use twice

khanalpride:
Two factor authentication code was previously being able to use twice. So when a user is logged in, if anyone knows password and current OTP, they will be allowed to login.

Now, with this PR, OTP will be only allowed one time. This does not affect recovery code and it will work as it is.

### DIFF
--- a/src/Contracts/TwoFactorAuthenticationProvider.php
+++ b/src/Contracts/TwoFactorAuthenticationProvider.php
@@ -30,12 +30,4 @@ interface TwoFactorAuthenticationProvider
      * @return int|bool
      */
     public function verify($secret, $code, $ts = null);
-
-    /**
-     * Provide current otp code based on secret.
-     *
-     * @param  string $secret
-     * @return string
-     */
-    public function getCurrentOtp($secret);
 }
--- a/src/Contracts/TwoFactorAuthenticationProvider.php
+++ b/src/Contracts/TwoFactorAuthenticationProvider.php
@@ -32,7 +32,7 @@ interface TwoFactorAuthenticationProvider
     public function verify($secret, $code, $ts = null);
 
     /**
-     * Provide current otp code based on secret
+     * Provide current otp code based on secret.
      *
      * @param  string $secret
      * @return string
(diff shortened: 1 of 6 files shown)


src/Contracts/TwoFactorAuthenticationProvider.php
```diff
@@ -26,7 +26,16 @@ public function qrCodeUrl($companyName, $companyEmail, $secret);
...
-    public function verify($secret, $code);
+    public function verify($secret, $code, $ts = null);
+
+    /**
+     * Provide current otp code based on secret.
+     *
+     * @param  string $secret
+     * @return string
+     */
+    public function getCurrentOtp($secret);
```
driesvints, line 40: The changes to this contract are breaking changes on a stable release.

khanalpride: that was only added to support tests.
I can use it without that function.

driesvints:
@khanalpride there's still breaking changes to existing method signature. I think it's best that this is sent to master instead.

khanalpride:
> @khanalpride there's still breaking changes to existing method signature. I think it's best that this is sent to master instead.

Is the addition of field in users table the breaking change you are mentioning?
If its then I will send to master.

driesvints:
@khanalpride the changes here: https://github.com/laravel/fortify/pull/43/files#diff-2dce9a6a75e3592b6ef5a41e6a48066aR32

and here: https://github.com/laravel/fortify/pull/43/files#diff-bf2fa3e05425dcf170ae6770e4ac2fbaR58

khanalpride:
> @khanalpride the changes here: https://github.com/laravel/fortify/pull/43/files#diff-2dce9a6a75e3592b6ef5a41e6a48066aR32
> 
> and here: https://github.com/laravel/fortify/pull/43/files#diff-bf2fa3e05425dcf170ae6770e4ac2fbaR58

I don't thing that's breaking change.

driesvints:
@khanalpride it definitely is: https://3v4l.org/Z3NT4

khanalpride:
> @khanalpride it definitely is: https://3v4l.org/Z3NT4

ok I will submit to master then.

PR #45, opened 2020-09-17
[2.x] OTP code now can be used only once

khanalpride:
Two factor authentication code was previously being able to use twice. So when a user is logged in, if anyone knows password and current OTP, they will be allowed to login.

Now, with this PR, OTP will be only allowed one time. This does not affect recovery code and it will work as it is.

I opened PR on #43 but was told it contains breaking changes, so I submitted to master.


taylorotwell:
No plans to change this right now.

Issue #47, opened 2020-09-17
Enter 2FA input token before save model


jetwes:
At the moment it is impossible to verify that the user activated 2fa in his device.
It should be mandatory that the user has to confirm the activation with an actual 2fa code.
This is not possible at the moment because in "Laravel\Fortify\Actions\EnableTwoFactorAuthentication" the database field is filled on activation. The generation of the secret has to be done before and the secret has to be an option in the action so it is possible to check a valid code before the secret is saved to the database.

related to [https://github.com/laravel/jetstream/issues/74](url)

driesvints:
Gonna mark this as a duplicate of https://github.com/laravel/jetstream/issues/74 to keep the discussion focused.

jetwes:
understood. But the implementation belongs to fortify in my opinion ;)

jetwes:
ok - will look into this tonight

PR #48, opened 2020-09-17
added optional $secret variable to EnableTwoFactorAuthentication.php

jetwes:
With this PR it is possible to enable TwoFactorAuthentication with an already generated secret
This is necessary if you want to check if the user successfully activated his device before you save the encrypted secret to the database. Logic for checking has to be in the front-end (e.g. jetstream).
Related to : https://github.com/laravel/fortify/issues/47
Related to: https://github.com/laravel/jetstream/issues/74


taylorotwell:
I will likely implement this myself in the future.

PR #51, opened 2020-09-18
[1.x] Add routes middleware option to configuration

leo95batista:
Hi everyone!. 

I have created this PR with the idea of ​​being able to provide the end user with a mechanism to modify which ```middleware``` Laravel Fortify is going to use in each of its routes.

Basically the user would just have to go to the config file: ```config/fortify.php``` and adjust the value of the ```middleware``` key according to their needs.

By default the value of this key is:

```
'middleware' => ['web'],
```

I have also added a short description to the option within the configuration file. Regards

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -4,6 +4,7 @@ use Laravel\Fortify\Features;
 
 return [
     'guard' => 'web',
+    'middleware' => ['web'],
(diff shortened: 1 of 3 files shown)


taylorotwell:
This already exists in Fortify so only the configuration stub change is needed.

PR #52, opened 2020-09-18
[1.x] Resend email verification after user update

trainman75:
When Email Verification is enabled, send a verification request to the new email address when a user updates their email address.
Optional commented line will null the email_verified_at value, which will force the user to verify the new address before doing anything else while logged in.


ManojKiranA:
If possible there must be a config option to enable the email re-verfication

PR #59, opened 2020-09-18
Inertia response fix

classjunky:
SimpleViewResponse is made to handle only Illuminate Views, so it can't recognize Innertia Response. I was unsure how do you handle dependencies with third party packages so I didnt mention anything related to Inertia, I just checked the response.


taylorotwell:
$this->view can be a string... call_user_func on that string will break Fortify.

PR #63, opened 2020-09-19
Added validation to the reset password page

ali-alharthi:
Validations for `password` and `password_confirmation` inputs were not present when resetting a password.


ali-alharthi:
This is a mistake I think the ResetUserPassword validate that! But the errors are no where to found on the front-end I'll do more digging.

PR #64, opened 2020-09-19
Removed unnecessary/unused variable

ali-alharthi:
I believe `, $password` was a typo or copied by mistake from the reset password documentation (here)[https://laravel.com/docs/8.x/passwords#resetting-the-password].


taylorotwell:
The password is really passed but it is just not used in this case.

PR #65, opened 2020-09-19
[1.x] Add option to force the password to have a special character.

leo95batista:
Hi! I have created this PR with the intention of adding in the password validation rules the option to require that the password contain or not special characters. 

Regards.

### DIFF
--- a/src/Rules/Password.php
+++ b/src/Rules/Password.php
@@ -78,34 +78,34 @@ class Password implements Rule
         }
 
         switch (1) {
-            case ($this->requireUppercase
+            case $this->requireUppercase
                 && ! $this->requireNumeric
-                && ! $this->requireSpecialCharacter):
+                && ! $this->requireSpecialCharacter:
                 return __('The :attribute must be at least '.$this->length.' characters and contain at least one uppercase character.');
 
-            case ($this->requireNumeric
+            case $this->requireNumeric
                 && ! $this->requireUppercase
-                && ! $this->requireSpecialCharacter):
+                && ! $this->requireSpecialCharacter:
                 return __('The :attribute must be at least '.$this->length.' characters and contain at least one number.');
 
-            case ($this->requireSpecialCharacter
+            case $this->requireSpecialCharacter
                 && !$this->requireUppercase
-                && !$this->requireNumeric):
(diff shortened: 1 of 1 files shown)


DarthShmev:
Looks good to me.

src/Rules/Password.php
```diff
@@ -66,14 +77,39 @@ public function message()
...
-            return __('The :attribute must be at least '.$this->length.' characters and contain at least one number.');
-        } elseif ($this->requireUppercase && $this->requireNumeric) {
-            return __('The :attribute must be at least '.$this->length.' characters and contain at least one uppercase character and number.');
-        } else {
-            return __('The :attribute must be at least '.$this->length.' characters.');
+        switch (1) {
+            case $this->requireUppercase
+                && ! $this->requireNumeric
+                && ! $this->requireSpecialCharacter:
+                return __('The :attribute must be at least '.$this->length.' characters and contain at least one uppercase character.');
```
buismaarten, line 84: It should probably changed to something like: `return __('The :attribute must be at least :length characters and contain at least one uppercase character.', ['length' => $this->length]);`

taylorotwell:
This PR has merge conflicts. Also why `switch(1)` instead of `switch(true)`?

leo95batista:
Hello, good morning everyone! , I am reviewing, and I see that the PR #70 was approved first before this PR #65. For that reason the conflict.  In PR #70, they changed the style in which the ```validation message was returned```. When creating my PR, I can change the style, but I did not do it to maintain the style that existed initially.


**Before**
```
return __('The :attribute must be at least '.$this->length.' characters and contain at least one uppercase character.');
```

**Now**
```
return __('The :attribute must be at least :length characters and contain at least one uppercase character.', ['length' => $this->length]);
```

I think ```switch(1)``` and ```switch(true)``` mean the same in that context

src/Rules/Password.php
leo95batista: Please check it now

buismaarten: @leo95batista looks better!

leo95batista:
@taylorotwell , do you think anything else needs to be changed before doing the merge?

PR #66, opened 2020-09-19
[1.x] Make use of the existing password validation rule

mercuryseries:
Little cleanup by using the existing password validation rule: https://laravel.com/docs/8.x/validation#rule-password


ali-alharthi:
I believe you should pass a guard with that rule to make sure it works correctly with Fortify since it uses whatever guard the user specify.
It is set to use `web` guard by default but that is optional. So I believe that is why @taylorotwell call the after method and used the password from the passed `$user`.
But maybe you can pass the current guard - didn't check yet.

taylorotwell:
You can change this in your app as needed.

PR #68, opened 2020-09-19
Fix flawed logic in the UpdateUserProfileInformation action

### LINKED ISSUE #67: UpdateUserProfileInformation Action unverifies user's email regardless of change
- Fortify Version: `1.4.1`
- Laravel Version: `8.0`
- PHP Version: `7.4.5`
- Database Driver & Version: `MySQL 8.0.19`

### Description:
When a user updates their profile information, the `UpdateUserProfileInformation` action checks to see if the new email in the request equals the user's existing email on record. Regardless of rather it does or not, the action sets the user's `email_verified_at` property to `null`. The expected behaviour would be to only nullify `email_verified_at` if the new email in the request does not equal the user's existing email on record.

### Steps To Reproduce:

- Setup the latest version of Fortify
- Submit a PUT request to `/user/profile-information` with any `email` and `name`.

DarthShmev:
The issue at hand has already been disclosed in #67 . In short, this PR prevents the package from forcing a user to reverify their email when simply updating their profile information (without changing their email).

Closes #67

### DIFF
--- a/stubs/UpdateUserProfileInformation.php
+++ b/stubs/UpdateUserProfileInformation.php
@@ -30,15 +30,18 @@ class UpdateUserProfileInformation implements UpdatesUserProfileInformation
             ],
         ])->validateWithBag('updateProfileInformation');
 
-        if ($input['email'] !== $user->email &&
-            $user instanceof MustVerifyEmail) {
+        if ($input['email'] !== $user->email && $user instanceof MustVerifyEmail) {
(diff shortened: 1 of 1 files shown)


SanderMuller:
Perhaps a test should be added that verifies the user only gets unverified on email change and a reverification notification gets sent when the email address is changed, so this does not break again in the future.

PR #76, opened 2020-09-21
Add LoginRateLimiter to $authenticateUsingCallback

nicolaferrara82:
In this case who have an active boolean field in User model or some custom functions for authentication  can increment $limiter inside FortifyServiceProvider with Fortify::authenticateUsing(); and use custom ValidationException message.


taylorotwell:
You can just resolve it out of the container if you need it.

Issue #73, opened 2020-09-21
Customize Password.php messages


jmpagella:
Recently i see the PR https://github.com/laravel/fortify/pull/70 to get a better way to translate those strings and I really appreciate that, but what if a want to custom that messages? Maybe a way to achieve that is with lang files or a possibility to change it in ValidationRules action published.

leo95batista:
Please, see mi PR #79 . Regards

jmpagella:
I just saw the PR and I think it is a good way to modify the text of those messages. I think Taylor thought it was just to translate and how he said, that function was applied in the PR that I mentioned earlier. Thanks anyway.
Please let me know if i close this issue

driesvints:
Closing because related PR was closed.

PR #79, opened 2020-09-23
[1.x] Option to customize translations

leo95batista:
Hello, as a result of having read issue #73, I have motivated myself to create a solution that can solve the problem. Basically this PR, adds the option to be able to customize the validation messages, and creates the bases to be able to make use of the translations if it is required in the future. 

Initially I have only created a file:
```
resources/lang/en/validation.php
```
that contains the validation messages that are required in the rules file:
```
src/Rules/Password.php
```

In this way, if the user wanted to modify the message returned by the validation, he would only have to publish the translations of the package in this way:

```
php artisan vendor:publish --tag fortify-translations
```
As a consequence of the result of this command, a file should be created in the application within the path:

```
resources/lang/vendor/fortify/en/validation.php
```

In this way, I think it is much more convenient for the user to be able to modify these messages. 

@taylorotwell I haven't really come up with a better name for the keys within the array of the file that contains the validation messages. I hope my idea can be accepted .. Regards!

### DIFF
--- a/resources/lang/en/validation.php
+++ b/resources/lang/en/validation.php
@@ -1,0 +1,24 @@
+<?php
+
+/*
+|--------------------------------------------------------------------------
+| Laravel Fortify Validation Language Lines
+|--------------------------------------------------------------------------
+|
+| The following language lines contain the default error messages used by
+| the validator class inside Laravel Fortify. Feel free to tweak each
+| of these messages here.
+|
+*/
+
+return [
+    'password' => [
+        'min' => 'The :attribute must be at least :length characters.',
+        'uppercase' => 'The :attribute must be at least :length characters and contain at least one uppercase character.',
+        'numeric' => 'The :attribute must be at least :length characters and contain at least one number.',
+        'special_character' => 'The :attribute must be at least :length characters and contain at least one special character.',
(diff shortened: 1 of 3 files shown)


taylorotwell:
You should already be able to customize the translation of these messages.

leo95batista:
Hello @taylorotwell , the truth is that I have not found how to do it. Issue #73 presents the following problem:

When registering a user, for example, how could the message returned by the validation be modified? 

Example:

- The password does not contain the minimum number of characters, by default 8.
- An error message will be thrown:
``` The :attribute must be at least :length characters. ```
https://github.com/laravel/fortify/blob/77bd46ac1729939a8db97b890a4d752dce79dc53/src/Rules/Password.php#L124

But, in case the user wanted to show this custom error, let's say they want to show it in Chinese language. How could it be modified?

That was really the goal that motivated me to create this PR. In order to make translations much more flexible.

jmpagella:
@leo95batista To translate an entire text string as a key you can access this link:
https://laravel.com/docs/8.x/localization#using-translation-strings-as-keys
What cannot be done and if it could with your PR is to modify the return in English using overwriting:
https://laravel.com/docs/8.x/localization#overriding-package-language-files

Issue #86, opened 2020-09-23
[Password Confirmation] Submitting an empty password field results in a server error


johanvanhelden:
- Fortify Version: 1.5.0
- Laravel Version: 8.6.0
- PHP Version: 7.4

During the implementation of the password confirmation functionality, I noticed the request is not validated beforehand to see if the required password is actually submitted. Resulting in a server error:

```
TypeError
Argument 3 passed to Laravel\Fortify\Actions\ConfirmPassword::__invoke() must be of the type string, null given, called in /var/www/projects/laravel/vendor/laravel/fortify/src/Http/Controllers/ConfirmablePasswordController.php on line 53 
```

If you submit an empty password field, the validation should just fail, as if you submitted an invalid password. It should not actually result in a server error.

PR #87, opened 2020-09-23
[1.x] Fix for empty password during confirmation

johanvanhelden:
Allow an empty password to be given and let it behave as an invalid password.

Before this fix, if the user leaves the field empty, a server error would occur. Now, a nice validation error will show.

---

Fixes https://github.com/laravel/fortify/issues/86


taylorotwell:
Does the other method in this class that type hints string password need that same change?

johanvanhelden:
@taylorotwell Yes, it does! I've added a fix and tests.

PR #90, opened 2020-09-25
[1.x] Respect the name of the user table in the action stub

tontonsb:


taylorotwell:
May be best to use the unique rule builder so you can just pass the model instance directly and have the table name extracted.

Issue #77, opened 2020-09-22
Redirection customization


webchopin:
It would be nice to offer customization options for redirection such as `Fortify::redirectAfterLoginUsing($callback)` or `Fortify::redirectAfterRegisterUsing($callback)`. This would bring back `laravel/ui` `redirectTo` capabilities. What do you think ? Do you prefer to encourage people to use custom Response classes and to put them under `app/Http/Responses/Fortify` ?

Concerning the registration use case, maybe it's pretty common to redirect users to a tour/guide page, so what do you think about a dedicated configuration key to set the path ?

rafayrty:
Hi There Add This To The Route File It Should Work.
```
Route::get('/login',function(){
    return redirect()->route('index');
})->name('login');
```

Adhik13th:
i think its not like that

nadhifikbarw:
Here's a workaround that if you need to implement it right now. I would absolutely prefer much more cleaner solution but for now this will do

---
Add new config key in `config/fortify.php`
```php
'registered' => '/onboarding',
```
or you could also make it use `RouteServiceProvider` const
```php
// app/providers/RouteServiceProvider.php
public const REGISTERED_REDIRECT = '/onboarding'


// app/config/fortify.php
'registered' => App\Provider\RouteServiceProvider::REGISTERED_REDIRECT,
```

Create new class `App\Responses\RegisterResponse`
```php
<?php

namespace App\Responses;

use Illuminate\Http\Response;
use Laravel\Fortify\Contracts\RegisterResponse as RegisterResponseContract;

class RegisterResponse implements RegisterResponseContract
{
    public function toResponse($request)
    {
        return $request->wantsJson()
            ? new Response('', 201)
            : redirect(config('fortify.registered'));
    }
}
```


Bind new interface implementation inside `FortifyServiceProvider.php`
```php
use App\Responses\RegisterResponse;
use Laravel\Fortify\Contracts\RegisterResponse as RegisterResponseContract;

/**
* Register any application services.
*
* @return void
*/
public function register()
{
    $this->app->bind(
        RegisterResponseContract::class,
        RegisterResponse::class
    );
}
```

jonasvanderhaegen:
Using Laravel v8.9.0, On top of RouteServiceProvider:

    /**
     * The path to the "home" route for your application.
     *
     * This is used by Laravel authentication to redirect users after login.
     *
     * @var string
     */
    public const HOME = '/';

joshmorel:
[...]

PR #93, opened 2020-09-28
Allow running code after the user is registered

m1guelpf:
This PR introduces a `created()` method similar to the old `registered()` method. It not only allows running code after the user has been created and logged in but also allows the user to return a custom response from this method that will override the default registration response.

The PR is targeted to `master` because updating contracts is a breaking change, which could be removed by using `method_exists` and not adding the method to the contract. Let me know if that's a preferred implementation.

### DIFF
--- a/src/Contracts/CreatesNewUsers.php
+++ b/src/Contracts/CreatesNewUsers.php
@@ -11,4 +11,12 @@ interface CreatesNewUsers
      * @return mixed
      */
     public function create(array $input);
+
+    /**
+     * The user has been created.
+     *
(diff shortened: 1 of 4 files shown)


taylorotwell:
I personally don't love tons of method hooks like this. I would suggest just moving the Registered event down a line so that it fires after login and people can put custom logic in an event listener that listens to that event.

m1guelpf:
Would you be open to a solution similar to what was proposed on #77?

driesvints:
@m1guelpf seems like Taylor indicated that he is on the issue

PR #97, opened 2020-09-29
[1.x] Update RegisteredUserController.php

lmhaydi:
Allow developpers to choose if Fortify will automatically login user after registration or not.
By default, the user will be logged in


taylorotwell:
No plans to add this for now.

Issue #100, opened 2020-09-29
ProfileInformationController status set to 'profile-information-updated'


thewebartisan7:
After succesfully update of profile information the status is set to 'profile-information-updated'

This string is not even possible to translate.

In my case I have in my main layout something like:

```
        @if (session('status'))
            {{ session('status') }}
        @endif
``` 

And can't even translate that string because the status can be already translated or have different namespace of translation.

Can we add a translated string there? Override that controller for just change that string seem not worth.

In Jetstream you seem just return a generic "Save" message from Livewire, but non-livewire and non-ajax submission need a way to display message from server.

Same thing for update password and two factors. Return 'password-updated' and 'two-factor-authentication-enabled'

Thanks

driesvints:
If you want to change anything here feel free to send in a PR. Thanks

thewebartisan7:
Done. English is not my mother tongue, so I would not be the right person to write these messages. I hope they go well.

PR #101, opened 2020-09-30
Change profile information updated status

thewebartisan7:
Ref https://github.com/laravel/fortify/issues/100

### DIFF
--- a/src/Http/Controllers/ProfileInformationController.php
+++ b/src/Http/Controllers/ProfileInformationController.php
@@ -23,6 +23,6 @@ class ProfileInformationController extends Controller
 
         return $request->wantsJson()
                     ? new JsonResponse('', 200)
-                    : back()->with('status', 'profile-information-updated');
+                    : back()->with('status', __('Profile information updated successfully'));
     }
 }


taylorotwell:
I see what you're saying but this is a breaking change and yes it is possible to translate? Just add a "verification-link-sent" entry to a "resources/lang/en/statuses.php" translation file and then display "@trans('statuses'.$status)"?

thewebartisan7:
@taylorotwell I think to do something like that, but there is already some sessions `status` already translated even in Fortify, see:

https://github.com/laravel/fortify/blob/1.x/src/Http/Responses/SuccessfulPasswordResetLinkRequestResponse.php#L38
https://github.com/laravel/fortify/blob/1.x/src/Http/Responses/PasswordResetResponse.php#L38

In old Laravel auth UI the statutes was already translated, and I used to do so also for my other responses. I could just change my own code that use status for response, but for this translated fortify response, I have to add additional check if translation is already translated.

I can't see that this "statutes" is something already used, so I suppose this translated phrase are translated via JSON?

PR #104, opened 2020-10-01
[1.x] Make TFA rememberable

### LINKED ISSUE #95: Option to remember 2FA code for x days
Would be nice to have an option to remember the 2fa code for e.g 7 days for a device.

Plesk for example is doing something similar (the days are configurable in this case): 
image

Chrissi2812:
It's based on the `ConfirmPassword` function from jetstream, it stores a timestamp in your session and later compares it.
Also a check for the user_id had to be implemented, as the session data for `2fa` will get restored after logout.

With this you can add a checkbox to the to the `two-factor-challenge.blade.php` inside jetstream and if it's checked the current device wont be bothered.

Currently if you log in as another user your remembered state would get overridden.

I'm not sure if this is the correct approach to this, so please let me know what you think.

Closes #95

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -20,4 +20,5 @@ return [
         Features::updatePasswords(),
         Features::twoFactorAuthentication(),
     ],
+    'two_factor_timeout' => 604800,
 ];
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #106, opened 2020-10-02
[1.x] Add FailedTwoFactorLoginResponse contract

faustbrian:


driesvints:
Can you give some explanation on why this is needed?

faustbrian:
Several applications I am working on have differing rules on how two-factor authentication failures are handled which diverge from the default of Fortify, which is to send the user back to the login, while I would prefer to show the user an error message.

```
<?php

namespace ARKEcosystem\Fortify\Responses;

use Illuminate\Contracts\Support\Responsable;
use Illuminate\Validation\ValidationException;

class FailedTwoFactorLoginResponse implements Responsable
{
    /**
     * Create an HTTP response that represents the object.
     *
     * @param \Illuminate\Http\Request $request
     *
     * @return \Symfony\Component\HttpFoundation\Response
     */
    public function toResponse($request)
    {
        $message = __('The provided two factor authentication code was invalid.');

        if ($request->wantsJson()) {
            throw ValidationException::withMessages([
                'code' => [$message],
            ]);
        }

        $request->session()->put([
            'login.id' => $request->session()->get('login.idFailure'),
        ]);

        return back()->withErrors(['code' => $message]);
    }
}
```

This would be an example of an alternative implementation to get rid of the redirect to the login page and also restoring the user ID to avoid the user being redirected to the login because `session('login.id')` would be empty.

PR #107, opened 2020-10-02
[1.x] Allow Fortify views to accept Responsable objects

claudiodekker:
This fix allow methods such as `Fortify::loginView` to return (amongst others) an Inertia response without running into this error: [Argument 1 passed to Symfony\Component\HttpFoundation\Response::setContent() must be of the type string or null, object given](https://flareapp.io/share/dmkdzMYP)

This happens because Laravel's router calls `->toResponse()` on `Responsable`, which SimpleViewResponse is an implementation of. After the object is converted to a response, other methods inside the router make sure that the response is properly generated.

However, because views such as Inertia's have their own Responsable implementation, this never gets triggered, because it never gets called by the SimpleViewResponse. This PR fixes this behaviour, preventing the user the need to do something like this:

```php
Fortify::loginView(function (Request $request) {
    return Inertia::render('Auth/Login')->toResponse($request);
});
```

### DIFF
--- a/src/Http/Responses/SimpleViewResponse.php
+++ b/src/Http/Responses/SimpleViewResponse.php
@@ -51,6 +51,7 @@ class SimpleViewResponse implements
         }
 
         $response = call_user_func($this->view, $request);
+
         if ($response instanceof Responsable) {
             return $response->toResponse($request);
         }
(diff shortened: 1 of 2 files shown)


reinink:
Nice work @claudiodekker, this is a welcome change for Inertia.js users! ❤️

PR #112, opened 2020-10-05
Per-environment fortify features

bredmor:
This change allows fortify features to be enabled/disabled via environment config, for example - enabling/disabling email verification with an `EMAIL_VERIFICATION` line in the `.env` file, e.g.:

```php
    'features' => [
        Features::registration(),
        Features::emailVerification([
            'enabled' => env('EMAIL_VERIFICATION', true)
    ]),
```

This gives the option for local/staging environments to remove barriers to testing without having to commit environment specific changes to VCS.

Some of these options I do not see this being particularly useful on, but I figured giving the option to all was better than trying to decide which would be potentially used. This change is fully backwards compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,18 @@ The `fortify` configuration file contains a `features` configuration array. This
 
 If you are not using Laravel Jetstream, you should implement user profile updates, password updates, and two-factor authentication yourself.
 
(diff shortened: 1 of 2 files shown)


driesvints:
Heya. The changes made to the method signatures are breaking changes. Please send to master for the next major release.

PR #116, opened 2020-10-06
Per-environment fortify features

bredmor:
Re-targeted #112 to `master`.


m1guelpf:
Couldn't you just do

```php
'features' => [
        Features::registration(),
        Features::resetPasswords(),
        env('EMAIL_VERIFICATION', true) ? Features::emailVerification() : null,
        Features::updatePasswords()
    ],
```

bredmor:
You're right, that would be quite a bit simpler wouldn't it?

PR #127, opened 2020-10-19
[1.x] Only use two factor action when enabled

stevebauman:
## Purpose

This PR resolves the issue of validating credentials twice when two-factor is disabled and the default login pipeline is used:

https://github.com/laravel/fortify/blob/c77b1832eed4223fa9e8fe1a5afc795c00c126a5/src/Http/Controllers/AuthenticatedSessionController.php#L83-L88

## Description

When two-factor authentication is disabled via the configuration file, users credentials are still validated twice, even though the action can be bypassed completely and another database call / `$authenticateUsing` callback execution can be avoided.

This is due to the `RedirectIfTwoFactorAuthenticatable` action, as well as the `AttemptToAuthenticate` action.

This was brought to light when overriding the `$authenticateUsing` callback, since it will be called twice, even when two-factor is omitted in `features`.


stevebauman:
Thank you! ❤️

PR #132, opened 2020-10-30
[1.x] Use "email" rule for default username

redidizzy:
A quick verification of username type in LoginRequest..
I started a brand new Laravel project using fortify, and having the email as the default username in a fresh laravel install, I noticed the validator isn't checking whether the email field is actually an email or not(It's using the string rule but it should be using the email rule if i'm using the default username chosen by Laravel)

### DIFF
--- a/src/Http/Requests/LoginRequest.php
+++ b/src/Http/Requests/LoginRequest.php
@@ -24,8 +24,10 @@ class LoginRequest extends FormRequest
      */
     public function rules()
     {
+        $username = Fortify::username();
+        $type = $username == 'email' ? 'email' : 'string'; 
         return [
-            Fortify::username() => 'required|string',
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure it actually matters? If the input is valid they will be logged in.

phpmac:
> 我不确定这真的很重要吗？如果输入有效，他们将被登录。

sir, did you known web3?so now,I use metamask. when login ,we don't need username and password,just like login using google.

phpmac:
> 我觉得这真的很重要吗？如果输入有效，他们将被登录。

Thank the team for all the efforts I hope to support it, because we can't do it now.
look this.

截屏2022-10-11 04 45 27
截屏2022-10-11 04 45 15

Issue #128, opened 2020-10-21
Having an option to enable only "api" routes


alexandrebordiere:
I think it would be nice to have, either a global configuration or a configuration at the feature level to enable only "api" routes.
I am using Fortify with a SPA so I don't need the views shipped with it and it's quite "annoying" when I run `php artisan route:list`.

I am ready to start the development if this feature is approved :smile:

Please, find examples below:

```php
/*
 |--------------------------------------------------------------------------
 | Disabling views 
 |--------------------------------------------------------------------------
 |
 | ...
 |
 */

  'disableViews' => false,
```
or

```php
'features' => [
    Features::registration(['disableViews' => true]),
],
```

alexandrebordiere:
I think this issue should be closed as the PR has been merged :smile:

alexandrebordiere:
I am closing the issue as the PR has been merged :smile:

PR #133, opened 2020-10-31
[1.x] Add `disable_views` configuration

alexandrebordiere:
This PR is the implementation for the following issue: https://github.com/laravel/fortify/issues/128

It allows the user to disable routes returning views if they are using their own front-end (SPA, etc...).


driesvints:
This was later changed to a "view" configuration which is enabled by default: https://github.com/laravel/fortify/commit/ff155d0e136d67bfd3832e052ed54135525c4569

PR #162, opened 2020-12-17
[2.x] Events for Two Factor Authentication

joelbutcher:
This PR adds three events to fortify that get fired in the given scenarios:

1. Two factor is enabled
2. Two factor is disabled
3. User generates new recovery code

I'm unsure in the last one whether or not to pass the recovery code to the event - can someone advise whether or not this would be useful?

Inspired by the [closed issue](https://github.com/laravel/jetstream/issues/313) on Laravel Jetstream requesting the logic be overwritten / extendable.

### DIFF
--- a/src/Actions/DisableTwoFactorAuthentication.php
+++ b/src/Actions/DisableTwoFactorAuthentication.php
@@ -2,6 +2,8 @@
 
 namespace Laravel\Fortify\Actions;
 
+use Laravel\Fortify\Events\TwoFactorDisabled;
+
 class DisableTwoFactorAuthentication
 {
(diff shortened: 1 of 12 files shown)


taylorotwell:
Can you name the events "TwoFactorAuthenticationDisabled" etc.

joelbutcher:
@taylorotwell done!

driesvints:
@joelbutcher @danclaytondev can any of you two send this PR and #239 to 1.x? It's unnecessary to do a new major release for those.

joelbutcher:
@driesvints Done :)

PR #164, opened 2020-12-17
[1.x] Improve Email Verification For SPA's using Laravel as an API

garethredfern:
Currently you are unable to use the VerifyEmailController with Fortify if you want to have an SPA which doesn't use Laravel's view templates. The reason for this is Laravel set's the route middleware to be `signed` and doesn't allow for using `signed:relative`. Also the VerifyEmailController currently only redirects to the home url set in the config template.

This PR enables a config variable for setting `signed:relative` but defaults to `signed` and returns json within the VerifyEmailController if the request wantsJson.

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -12,6 +12,7 @@ return [
     'home' => '/home',
     'prefix' => '',
     'domain' => null,
+    'signed' => 'signed',
     'limiters' => [
         'login' => null,
     ],
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -80,8 +80,10 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
                 ->name('verification.notice');
         }
 
+        $signed = config('fortify.signed');
+
         Route::get('/email/verify/{id}/{hash}', [VerifyEmailController::class, '__invoke'])
-            ->middleware(['auth', 'signed', 'throttle:6,1'])
+            ->middleware(['auth', $signed ? $signed : 'signed', 'throttle:6,1'])
             ->name('verification.verify');
 
         Route::post('/email/verification-notification', [EmailVerificationNotificationController::class, 'store'])
--- a/src/Http/Controllers/VerifyEmailController.php
+++ b/src/Http/Controllers/VerifyEmailController.php
@@ -3,7 +3,7 @@
 namespace Laravel\Fortify\Http\Controllers;
 
 use Illuminate\Auth\Events\Verified;
-use Illuminate\Http\RedirectResponse;
+use Illuminate\Http\JsonResponse;
 use Illuminate\Routing\Controller;
 use Laravel\Fortify\Http\Requests\VerifyEmailRequest;
 
@@ -13,18 +13,22 @@ class VerifyEmailController extends Controller
      * Mark the authenticated user's email address as verified.
      *
      * @param  \Laravel\Fortify\Http\Requests\VerifyEmailRequest  $request
-     * @return \Illuminate\Http\RedirectResponse
+     * @return \Illuminate\Http\RedirectResponse|\Illuminate\Http\JsonResponse
      */
-    public function __invoke(VerifyEmailRequest $request): RedirectResponse
+    public function __invoke(VerifyEmailRequest $request)
     {
         if ($request->user()->hasVerifiedEmail()) {
-            return redirect()->intended(config('fortify.home').'?verified=1');
+            return $request->wantsJson()
+                ? new JsonResponse('', 204)
+                : redirect()->intended(config('fortify.home').'?verified=1');
         }
 
         if ($request->user()->markEmailAsVerified()) {
             event(new Verified($request->user()));
         }
 
-        return redirect()->intended(config('fortify.home').'?verified=1');
+        return $request->wantsJson()
+            ? new JsonResponse('', 204)
+            : redirect()->intended(config('fortify.home').'?verified=1');
     }
 }
--- a/stubs/fortify.php
+++ b/stubs/fortify.php
@@ -80,6 +80,17 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Fortify Verify Email
+    |--------------------------------------------------------------------------
+    |
+    | Here you may specify whether Fortify will assign the Email Verification
+    | route signed or signed:relative for validating relative urls.
+    */
+
+    'signed' => 'signed',
+
+    /*
+    |--------------------------------------------------------------------------
     | Fortify Routes Middleware
     |--------------------------------------------------------------------------
     |


driesvints:
Please resubmit this when this is ready, thanks.

garethredfern:
Hi @driesvints this is my first PR. Would you suggest that I need to write a test for this? Other than that it's ready. Any advice would be appreciated.

driesvints:
@garethredfern ah you submitted this as a draft so it didn't seem like it was finished yet.

garethredfern:
Ok, apologies for that. Shall I resubmit as it is. Like I say I'm happy with it and all existing tests pass. I just wasn't sure if I **had** to write a test as it's a fairly small change?

driesvints:
Thanks, Taylor will review this once he gets to it.

taylorotwell:
So walk me through the flow of how you picture this working. Without this change, normally I would expect the user to click the button in the email to verify their email address which takes them to a URL on your backend Laravel API, which thens redirects them to your SPA.

garethredfern:
A user will receive a verify email either on registration or if they click on a verify email button in the SPA. This then sends an email as you would normally expect by hitting: `Laravel\Fortify\Http\Controllers\EmailVerificationNotificationController@store`.

However, in the `AuthServiceProvider` I need to add the following to redirect to the SPA with a valid `id`, `hash`, `signature` etc. My preference is to not redirected to the Laravel API so they always see the same URL for the front end and only interact with the SPA.

```php
VerifyEmail::createUrlUsing(
  function ($notifiable) {
    $url = URL::temporarySignedRoute(
      'verification.verify',
      Carbon::now()->addMinutes(Config::get('auth.verification.expire', 60)),
      [
        'id'   => $notifiable->getKey(),
        'hash' => sha1($notifiable->getEmailForVerification()),
      ],
      false // this will disable the absolute route & return a relative route
    );

    return "http://localhost:8080{$url}"; // this is the SPA url
  }
);
```

The SPA can grab the returned data (user must be logged in) and do a get request back to the Laravel API.

```js
async verifyEmail() {
  const id = this.$route.params.id;
  const hash = this.$route.params.hash;
  const expires = this.$route.query.expires;
  const signature = this.$route.query.signature;
  await axios.get(
    `http://api.laravel.test/email/verify/${id}/${hash}?expires=${expires}&signature=${signature}`
  );
},
```

When the `GET` request above is made it hits: `Laravel\Fortify\Http\Controllers\VerifyEmailController@__invoke`

To be able to get the signed middleware to work I needed the `verification.verify` route to be changed from `signed` to the `signed:relative` option. That is my first change. In the fortify routes file.

Once the signed middleware passes we then move into the `VerifyEmailController`.

In this controller I am returned to the home view of the Laravel API unless I change the `home` variable in the fortify config which redirects me to `'http://localhost:8080/?verified=1‘`. If I do this, in my console, I can see a 304 response with the response payload of a Laravel view. My SPA URL doesn’t change to `http://localhost:8080/?verified=1` maybe I am misunderstanding something?

Screenshot 2020-12-18 at 15 42 10

With the changes I have made to the `VerifyEmailController` I would receive a 204 response and could you could send a json message saying `{ verified: true }`. This can then be used in the SPA to trigger a email verified message.

sambindoff:
@garethredfern I think you're overcomplicating things by having the SPA handle verifying the email through an extra AJAX call. You can just let Fortify handle both sending the verify email notification and verifying the email address when the link is clicked, this works out the box even with a SPA.

I've created an open source package for SPA authentication with Fortify if you want to check it out as a guide [https://github.com/sambindoff/laravel-gust](https://github.com/sambindoff/laravel-gust).

Hope you don't mind me jumping in @taylorotwell, might help clear things up.

Cheers
Sam

garethredfern:
Thank you for the input @sambindoff. Fortify does everything I need to run an SPA without the need for any third party packages. My one sticking point is the email verification.

I must be missing something because on the latest install of Laravel, Fortify and Sanctum I get redirected to an error page which says `Route [login] not defined` if I follow these [instructions](https://laravel.com/docs/8.x/fortify#email-verification).

Which is why I then need to add the following in my AuthServiceProvider:

```php
VerifyEmail::createUrlUsing(
      function ($notifiable) {
        $url = URL::temporarySignedRoute(
          'verification.verify',
          Carbon::now()->addMinutes(Config::get('auth.verification.expire', 60)),
          [
            'id'   => $notifiable->getKey(),
            'hash' => sha1($notifiable->getEmailForVerification()),
          ],
          false // this will disable the absolute route & return a relative route
        );

        return "http://localhost:8080{$url}";
      }
    );
```

And to make the above work with Fortify out of the box you need the changes I have PR'd (or something similar).

sambindoff:
@garethredfern the package was more of a reference so you could see the changes required to make your SPA compatible with Fortify.

The issue has nothing to do with the `VerifyEmail::createUrlUsing` callback so you don't need to worry about overriding this in your provider. The `Route [login] not defined` error will be thrown from [https://github.com/laravel/laravel/blob/8.x/app/Http/Middleware/Authenticate.php#L18](https://github.com/laravel/laravel/blob/8.x/app/Http/Middleware/Authenticate.php#L18).

With Fortify, if you've changed `views => false` in `config/fortify.php`, there is no longer a named route for `/login` which is what the above is looking for. The only way around this is to change the reference from `route('login')` to `url('login')` in the above file.

Alternatively, you can add your own `login` named route to your `routes/web.php` which will also work, but then you'll have an unnecessary route defined.

Hope this helps,
Sam

garethredfern:
Thank you @sambindoff what you suggest does indeed work without any modification to to the AuthServiceProvider or Fortify.

The only thing I would say is, what if for some reason you do want to still have a sepearte login to the API which is nothing to do with the the SPA e.g. for admins or something. Doing what you suggest then breaks the normal flow for the Laravel APP. It would be nice to have Fortify handle the requirments for sperating out the SPA authentication.

I'll wait to see what Taylor's thoughts are but I would be happy going with what you suggest and closing this if that is the preferred approach.

taylorotwell:
@sambindoff @garethredfern so, in summary the only change needed to make this work is to change the redirect URL in your Authenticate middleware to redirect users to your SPA login page?

garethredfern:
Yes that is correct, but it does force you to redirect to an external url which is fine for my use case. The PR I submitted maybe adding a bit more complication to the SPA but it doesn't force a change to that login redirect. I'm happy to close this down though using @sambindoff solution as it's an easy solution for what I need.

sambindoff:
Yeh seems like a userland issue and not something Fortify need worry about.

@garethredfern not sure I fully follow, but if you need some form of acl with additional authentication requirements for admins with a separate login flow, that's something your application needs to implement. You could always hit the same Fortify endpoints to authenticate and then perform additional logic on the client. But maybe I'm not grasping exactly what you're trying to do.

garethredfern:
I am going to close this down as Sam's solution works well for me, @sambindoff thank you for taking the time to help and sharing your knowledge.

Issue #163, opened 2020-12-17
Constructor dependency injector avoids Laravel Fortify to work out of the box in Tenancy for Laravel package


acacha:
According to: 

https://tenancyforlaravel.com/docs/v3/early-identification

Using constructor DI avoids tenancy route middleware to works. AFAIK the proposed changes like using method injection instead of constructor will solve the issue.

I can write a Pull Request if your think the change is viable. Feel free to close this issue for any reason.

driesvints:
If you want to change anything just PR it, thanks

acacha:
PR created at https://github.com/laravel/fortify/pull/165

acacha:
PR created at https://github.com/laravel/fortify/pull/169  . Sorry for the multiple PR with errors.

PR #165, opened 2020-12-18
Use method DI instead of constructor DI to allow packages like Laravel tenancy works with Fortify Out of the box.

acacha:
As you can read at https://github.com/laravel/fortify/issues/163

According to:

https://tenancyforlaravel.com/docs/v3/early-identification

Using constructor DI avoids tenancy route middleware to works. AFAIK the proposed changes like using method injection instead of constructor will solve the issue.

I can write a Pull Request if your think the change is viable. Feel free to close this issue for any reason.


acacha:
Sorry error in tests. I will try again...

PR #169, opened 2020-12-18
Use method DI instead of constructor DI to allow packages like Laravel tenancy works with Fortify out of the box

acacha:
As you can read at #163

According to:

https://tenancyforlaravel.com/docs/v3/early-identification

Using constructor DI avoids tenancy route middleware to works. The reason is Constructor DI is executed before middleware.

AFAIK the proposed changes will solve the issue and will make building web apps with Laravel Tenancy package and Fortify/Jetstream easier but feel free to close this PR for any reason...

### DIFF
--- a/src/Http/Controllers/AuthenticatedSessionController.php
+++ b/src/Http/Controllers/AuthenticatedSessionController.php
@@ -78,7 +78,7 @@ class AuthenticatedSessionController extends Controller
      * @param  \Illuminate\Contracts\Auth\StatefulGuard $guard
      * @return \Laravel\Fortify\Contracts\LogoutResponse
      */
-    public function destroy(Request $request,StatefulGuard $guard): LogoutResponse
+    public function destroy(Request $request, StatefulGuard $guard): LogoutResponse
     {
         $guard->logout();
 
--- a/src/Http/Controllers/AuthenticatedSessionController.php
+++ b/src/Http/Controllers/AuthenticatedSessionController.php
@@ -20,24 +20,6 @@ use Laravel\Fortify\Http\Requests\LoginRequest;
 class AuthenticatedSessionController extends Controller
 {
     /**
-     * The guard implementation.
-     *
-     * @var \Illuminate\Contracts\Auth\StatefulGuard
-     */
-    protected $guard;
-
-    /**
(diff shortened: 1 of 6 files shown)


driesvints:
There are multiple breaking changes in this PR. It's unlikely we'll ever merge this sorry.

stancl:
I wanted to say that ideally there wouldn't have to be a PR to Fortify for this. It's tenancy that should be adjusted, since tenancy is what's 3rd party here.

But this PR seems pretty simple. @driesvints is there any way users could make a change like this from userland? i.e. change the controllers used by Fortify

driesvints:
@stancl you can use `Fortify::ignoreRoutes()` and register your own routes.

stancl:
I see. I just thought of an interesting solution.

To explain the issue simply, my package finds the tenant using the request data and then it switches the application context to the tenant's context for the request. 

However, controller constructors are executed before middleware (since they can specify middleware) and tenancy is initialized in a middleware. 

So any dependencies injected in the controller constructor will use the central context.

Now it occurred to me that this could be solved by a proxy controller. Either defer the real controller instantiation, or replace dependencies. The former solution is cleaner, but it removes the ability to specify middleware in a constructor, for example.

I'll play with this, could be a good feature for compatibility. 

Thanks @driesvints!

driesvints:
I see your problem. In that regard the changes in this PR could indeed definitely help. You can always try it against master to see if Taylor would accept it.

acacha:
Thanks @driesvints and @stancl for you time! And again sorry for the multiple PRs... 

I will submit the PR again to master asap.

driesvints:
@acacha don't worry about it 👍

stancl:
Yeah, I basically recommend people to use route action DI rather than constructor DI. https://tenancyforlaravel.com/docs/v3/early-identification/

It works for most things & all userland code, but Fortify has been the #1 thing my users have been struggling with recently.

As I said, I'd prefer if my package could integrate with other packages seamlessly, rather than adding band aids to all packages in the ecosystem. 

So I'll attempt a solution such as the deferred controller proxy I mentioned. And only if that fails I'll propose a change for the next major release.

PR #173, opened 2020-12-21
Auto enable EmailVerification

AnandPilania:
Auto enable EmailVerification feature when MAIL_params is configured in .env

### DIFF
--- a/stubs/FortifyServiceProvider.php
+++ b/stubs/FortifyServiceProvider.php
@@ -18,7 +18,9 @@ class FortifyServiceProvider extends ServiceProvider
      */
     public function register()
     {
-        //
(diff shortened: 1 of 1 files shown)


stubs/FortifyServiceProvider.php
```diff
@@ -18,7 +18,9 @@ class FortifyServiceProvider extends ServiceProvider
      */
     public function register()
     {
-        //
+        if (env('MAIL_USERNAME', null) && env('MAIL_PASSWORD', null)) {
```
driesvints, line 21: You can't use the `env` helper here directly as it won't read ENV values when config is cached.

PR #176, opened 2020-12-23
[1.x] Added guard routes file

thecaliskan:
Hi;

I use this system with Nova. Two guards are required in this scenario. If there is a possibility of define, it should be. No time should be wasted to find.

Needs to be standard in config files on Laravel Official Projects.

Thanks


taylorotwell:
Why are two guards required? Can you elaborate?

thecaliskan:
Hi;

Admin(web) guard used in Nova. Client guard used in Jetstream.

taylorotwell:
Let the Jetstream guard be your default guard.

thecaliskan:
Guard system is not third party library. Why don't you want to support ?

Issue #171, opened 2020-12-21
Is TwoFactorAuthenticatedSessionController missing PrepareAuthenticatedSession?


LeoniePhiline:
- Fortify Version: 1.7.2
- Laravel Version: 8.19
- PHP Version: 7.4

#### Authentication with vs. without 2fa
When authenticating **without** the `Features::twoFactorAuthentication()` feature then `AuthenticatedSessionController` pipes `$request` through `PrepareAuthenticatedSession`. If **2fa** is used, though, then this step is skipped and we continue in `TwoFactorAuthenticatedSessionController`.
The latter still performs a `$session->migrate(true)` through `$guard->login(...)` (like `AttemptToAuthenticate` for non-2fa logins), but does not regenerate the **CSRF token** (`$session->regenerate()`), like `PrepareAuthenticatedSession` would perform for non-2fa logins. Neither is the **limiter cleared** after successful 2fa login. These points made me wonder if `TwoFactorAuthenticatedSessionController` was missing a continuation of `AuthenticatedSessionController`'s login pipeline.

#### Missing login rate limiter increment at failed 2fa challenge
When failing to enter the correct 2fa code, `LoginRateLimiter::increment()` is not called - only if the initial password challenge fails. Doesn't this make brute-forcing the second factor possible?

#### Using existing user provider methods
On a side note, I noticed that `RedirectIfTwoFactorAuthenticatable::validateCredentials()` performs its own credentials check, where `AttemptToAuthenticate::handle()` uses the provider's `retrieveByCredentials()` method. It might make sense to use `\Illuminate\Contracts\Auth\UserProvider::retrieveByCredentials()` and `\Illuminate\Contracts\Auth\UserProvider::validateCredentials()` in `RedirectIfTwoFactorAuthenticatable` as well, instead of the current custom ad-hoc implementations, even if Fortify is meant to depend on Eloquent as users provider.

driesvints:
Thanks for the clear issue. I'll let Taylor check in on these.

driesvints:
@LeoniePhiline I think I got all of them. Taylor is going to review the PRs I made.

driesvints:
I'm going to close this now that we merged the main PR for this.

jandreasn:
@driesvints The rate limiter still doesn't seem to be cleared after two factor login. This line is missing from the two factor login flow:
https://github.com/laravel/fortify/blob/master/src/Actions/PrepareAuthenticatedSession.php#L38

[...]

PR #179, opened 2020-12-29
[1.x] Rate limit two factor auth

driesvints:
This is a partial fix for https://github.com/laravel/fortify/issues/171. At the moment the two factor challenge isn't rate limited, leaving it open to brute force attacks. 

I wasn't sure if the rate limit increment in the `FailedTwoFactorLoginResponse` is the correct location. @taylorotwell is this fine or do you think it's better placed somewhere else?

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -1,6 +1,7 @@
 <?php
 
 use Illuminate\Support\Facades\Route;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Pushed some fixes to address this... https://github.com/laravel/fortify/commit/8609af2292652234a70e4457d63ff1e10a510631

PR #180, opened 2020-12-29
[1.x] Retrieve user through provider

driesvints:
This is a partially fix for https://github.com/laravel/fortify/issues/171. Instead of relying on Eloquent to retrieve the user we make use of the user provider from the stateful guard and thus decouple the Two Factor implementation from Eloquent.

### DIFF
--- a/src/Actions/RedirectIfTwoFactorAuthenticatable.php
+++ b/src/Actions/RedirectIfTwoFactorAuthenticatable.php
@@ -4,7 +4,6 @@ namespace Laravel\Fortify\Actions;
 
 use Illuminate\Auth\Events\Failed;
 use Illuminate\Contracts\Auth\StatefulGuard;
-use Illuminate\Support\Facades\Hash;
 use Illuminate\Validation\ValidationException;
 use Laravel\Fortify\Fortify;
 use Laravel\Fortify\LoginRateLimiter;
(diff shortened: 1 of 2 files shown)


LeoniePhiline:
Hi @driesvints 

> ... and thus decouple the Two Factor implementation from Eloquent.

The handling of recovery keys is still tightly coupled to Eloquent (`$user->forceFill([...])->save()`):

- https://github.com/laravel/fortify/blob/13a7b0a72f6723384e5217ddfbca32167c3756bd/src/Actions/EnableTwoFactorAuthentication.php
- https://github.com/laravel/fortify/blob/41025cdb3836004775cfc7d8ba473dcfbdbe353e/src/Actions/GenerateNewRecoveryCodes.php
- https://github.com/laravel/fortify/blob/13a7b0a72f6723384e5217ddfbca32167c3756bd/src/Actions/DisableTwoFactorAuthentication.php
- https://github.com/laravel/fortify/blob/13a7b0a72f6723384e5217ddfbca32167c3756bd/src/TwoFactorAuthenticatable.php

Maybe that's entirely okay (I have never *not* used Eloquent) - only the claim that 2fa handling was now User-Provider-agnostic appears false to me. 

I'm unsure how this could be solved the most easily, since the Eloquent methods (`forceFill`/`save`) used on the user model are not defined in any interface.
 
The above-listed `Actions` would need a replaceable `EloquentAdapter` implementing a fill/fill-and-save interface. Any other ideas?

Leonie

driesvints:
@LeoniePhiline I'm with vacation as of just now. I'll try to look into this somewhere next week.

LeoniePhiline:
@driesvints It's not urgent at all. Have a wonderful time! 🎄❄️✨

taylorotwell:
Personally I don't have any problem with this package requiring Eloquent. I don't really have the desire at this point to make it support anything else.

LeoniePhiline:
@taylorotwell

Clarification, as the discussion might have been misleading:

This PR was only marginally about independence from Eloquent.

It was about leaving credentials work up to the `Provider` instead of assuming (hard-coding) a specific way to retrieve users and verify credentials _and_ duplicating this functionality's implementation.

The current implementation 
1. duplicates this already existing functionality and 
2. clashes with the ability to use of custom user providers.

Even if you don't care about 2.), you might still want to merge this PR to have 1.) fixed, to at least not re-implement the provider's functionality.

Developers might use a custom provider but still use Eloquent.

driesvints:
@LeoniePhiline I've resent it here: https://github.com/laravel/fortify/pull/189

PR #181, opened 2020-12-29
[1.x] Fix Two Factor prepare auth session

driesvints:
This is a partial fix for https://github.com/laravel/fortify/issues/171. After a successful two factor auth, the `PrepareAuthenticatedSession` action is never reached and thus a new CSRF token is never generated nor is the rate limiter reset. This PR adds these calls manually to the `TwoFactorAuthenticatedSessionController`. Ideally the `PrepareAuthenticatedSession` action would handle this but it can never be reached because two factor is being done in a new request.

There's also a few minor DocBlock fixes I sneaked in.

### DIFF
--- a/src/Http/Controllers/AuthenticatedSessionController.php
+++ b/src/Http/Controllers/AuthenticatedSessionController.php
@@ -29,7 +29,7 @@ class AuthenticatedSessionController extends Controller
     /**
      * Create a new controller instance.
      *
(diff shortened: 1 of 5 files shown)


taylorotwell:
Just pushed improvements to Fortify's rate limiting setup that handles this.

taylorotwell:
Sorry wrong PR.

PR #182, opened 2020-12-29
Change users table for unique validation

sifonsecac:
In case you use another table name or model rather than the users table to store users, the unique validation fails because it searchs in a non existing table. Instead, using its classname in the validation the application will determine correctly the table name to use.


taylorotwell:
Just update your stub to be what you want.

Issue #161, opened 2020-12-11
Using any delete on a model or DB::statement, DB::delete makes using Fortify::authenticateUsing fails.


EngHell:
So i have created a custom controller to hangle the "/login" post route, which is exactly as the one provided in the src but with a modified Request injection for renaming password to token.

The issue arises in this code(JetstreamServiceProvider.php):
```php
Fortify::authenticateUsing(function(Request $request){
            $token = LoginToken::
                with("user")
                ->where("user_email",$request->email)
                ->where("token",$request->token)
                ->first();



            if($token){
                $user = $token->user;
                /*
                LoginToken::where("user_email",$request->email)
                    ->delete();
                */

                DB::statement("DELETE FROM login_tokens where user_email=:email ",["email"=>$token->user_email]);

                //DB::raw("DELETE FROM login_tokens where user_email='".$token->user_email."';");
                //dd($user);
                return $user;
            }

        });
```

As you can see i have commented some code sections as i have been experimenting, even retrieving the user after the token delete with `User::where("email",$request->email)` and even checking with `dd` says `$user` is not null in all cases but even then the response in the default jetstream view is
> These credentials do not match our records.

But removing any delete query makes it work as intended.

Whats supposed to happen:
1. Token is searched in database
2. If token is found the user is saved to an $user variable
3. Then the token is deleted
4. Return the user and allow log in.

What happens:
At step 4 the log in fails even with the return not being null or false.

EngHell:
I've digging more inside of AttempToAuthenticate.php
```php
protected function handleUsingCustomCallback($request, $next)
    {
        $user = call_user_func(Fortify::$authenticateUsingCallback, $request);

        dd($user);

        if (! $user) {
            $this->fireFailedEvent($request);

            return $this->throwFailedAuthenticationException($request);
        }

        $this->guard->login($user, $request->filled('remember'));

        return $next($request);
    }
```

`$user` is null only after running any delete statement.

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
[...]

PR #185, opened 2021-01-02
Added a callback for RedirectIfTwoFactorAuthenticatable

EngHell:
Fully backward compatible, if no specific callback is set the normal callback will be used if it applies.

Inspired in my usecase where I had to delete a token at search but the callback is called twice so the second callback misses and fails, therefore instead of declaring a custom pipeline and creating a new extra class, just to do a separated step for deleting or any one time step that needs to be done while querying the credentials.

#161

### DIFF
--- a/src/Actions/RedirectIfTwoFactorAuthenticatable.php
+++ b/src/Actions/RedirectIfTwoFactorAuthenticatable.php
@@ -66,8 +66,10 @@ class RedirectIfTwoFactorAuthenticatable
      */
     protected function validateCredentials($request)
     {
-        if (Fortify::$authenticateUsingCallback) {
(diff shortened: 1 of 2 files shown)


taylorotwell:

EngHell:
At least the documentation should clearly state the the default pipeline calls twice the normal callback when two factor auth is enabled, so other people do not invest an hour of code debuging to realize they should modify the login pipeline just to add steps that cant be invoked twice in the data, which makes first, code duplication of the normal pipeline as what I think everyone would do is to copy the pipeline setup you have in fortify files, then register the same pipeline plus 1 step, which can be avoided, as well the pipeline can get outdated as you may change your default pipeline and developers trying to keep their pipeline as close to the default fortify pipeline will have to update it, instead of using an extra callback which requires less code.

PR #187, opened 2021-01-02
Redirect after register

victoralbino:
Implemented the possibility to perform some action after the user registers

Now it is only necessary to implement the afterRegister method in the App \ Actions \ Fortify \ CreatesNewUsers Class and what is in it will be executed after registration.

The goal was to be able to do some validation or even redirection in some cases, I also noticed that issue #77  has a similar need.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea/
 /vendor
 composer.lock
 /phpunit.xml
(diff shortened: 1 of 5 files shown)


taylorotwell:
See `Registered` event.

victoralbino:
For my case the user should lose the login and be redirected, I ended up using this solution.

Because after the event the user logs in.

If you're making a beginner's mistake, sorry for the pull request

PR #188, opened 2021-01-02
Remove 'guest' middleware from reset pwd routes

slopdo:
This PR removes 'guest' middleware from reset password routes (same change as @driesvints did in https://github.com/laravel/laravel/pull/5129) so that users can reset their password while logged in.


taylorotwell:
No plans to make this change.

PR #189, opened 2021-01-04
[1.x] Retrieve user through provider

driesvints:
Re-send for https://github.com/laravel/fortify/pull/180

### DIFF
--- a/src/Actions/RedirectIfTwoFactorAuthenticatable.php
+++ b/src/Actions/RedirectIfTwoFactorAuthenticatable.php
@@ -4,7 +4,6 @@ namespace Laravel\Fortify\Actions;
 
 use Illuminate\Auth\Events\Failed;
 use Illuminate\Contracts\Auth\StatefulGuard;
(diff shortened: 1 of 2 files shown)


src/Actions/RedirectIfTwoFactorAuthenticatable.php
```diff
@@ -76,15 +75,17 @@ protected function validateCredentials($request)
...
+        if (! $user) {
+            $this->fireFailedEvent($request, $user);
 
-                $this->throwFailedAuthenticationException($request);
-            }
-        });
+            $this->throwFailedAuthenticationException($request);
+        }
+
+        return $user;
```
mpyw, line 88: The verification might be not enough. `retrieveByCredentials()` does not expect password matches, only an email address does. Password hash validation is done by `validateCredentials()`, but its call was not included in the PR.

I think it is the cause of the PR reverted. 

@taylorotwell Is it right?

PR #198, opened 2021-01-14
[1.x] Custom login and two factor login rules.

bangbangda:
Custom rules, may be more convenient to add validation rules.
eg authentication code validation.

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -29,11 +29,10 @@ return [
         'login' => [
             Fortify::username() => 'required|string',
             'password' => 'required|string',
-        ], 
+        ],
         'twoFactorLogin' => [
             'code' => 'nullable|string',
             'recovery_code' => 'nullable|string',
-
         ],
     ],
 ];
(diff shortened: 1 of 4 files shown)


joelbutcher:
This would be better done via binding an implementation to a contract, then setting it via the Fortify class and calling it within the published service provider. Similar to the way fortify registers views fir its routes.

driesvints:
Since we're not providing defaults in the implementing classes themselves this will be a breaking change for anyone already implementing Fortify.

taylorotwell:

kukudeliu:
update: set login of custom verification rules.
reason: usually,verify the verification code.It will be more flexible.
@taylorotwell

kukudeliu:
> Since we're not providing defaults in the implementing classes themselves this will be a breaking change for anyone already implementing Fortify.

It will not affect already implementing

driesvints:
@kukudeliu it most definitely will for the people who have published the config file.

joelbutcher:
@bangbangda @taylorotwell I've submitted a new PR for this which should avoid the breaking change and allow developers to specify the rules used for validating login & two-factor authentication requests. #199

bangbangda:
Good idea, custom rules are necessary.
@joelbutcher

PR #199, opened 2021-01-15
[1.x] Allow validation rules to be overridden.

joelbutcher:
This PR is an attempt to allow developers to customise the rules used for authenticating login requests and two-factor authentication requests. It tries to avoid the breaking change that https://github.com/laravel/fortify/pull/198 presented.

Developers can override the rules by passing an array or a callback into the new methods in the `boot` method of the `FortifyServiceProvider`:

```php
/**
 * Bootstrap any application services.
 *
 * @return void
 */
public function boot()
{
    Fortify::validateLoginRequestsUsing([
        //
    ]);

    Fortify::validateLoginRequestsUsing(function () {
        return [
            //
        ];
    });

    Fortify::validateTwoFactorAuthenticationRequestsUsing([
        //
    ]);

    Fortify::validateTwoFactorAuthenticationRequestsUsing(function () {
        return [
            //
        ];
    });

    //...
}
```

### DIFF
--- a/src/Fortify.php
+++ b/src/Fortify.php
@@ -53,6 +53,13 @@ class Fortify
     public static $loginRules = null;
 
     /**
+     * The rules responsible for validating two-factor authentication requests.
+     *
+     * @var callable|array
+     */
+    public static $twoFactorAuthenticationRules = null;
+
+    /**
      * Get the username used for authentication.
      *
      * @return string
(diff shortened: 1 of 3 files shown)


taylorotwell:
I don't see a big reason you would need custom validation rules. If the input retrieves a valid user it is valid input. If the authentication code passes... it is valid. That *is* the validation.

srdante:
@taylorotwell I think that would be valid, because I was trying to put a captcha on my website, and the only way was to create middleware that would check which url was being called to apply the customized validation rule.

LUGAMAFE:
Man could you reconsider the possibility of adding our own validation logic, there are things like the ones that srdante says like validating a captcha that need to be validated. And by the time Fortify :: authenticateUsing () is reached, it is too late to validate.

joelbutcher:
@taylorotwell taking into account the above arguements, would you re-consider this PR?

joelbutcher:
@taylorotwell pruning some stale repo’s and branches and brought me back to this…

Given the above comments, I feel this is a valid PR for folks to be able to provide additional functionality to these routes, such as ReCAPCHA as mentioned above. 

Another example could be password hash validation. Someone may want to validate that the password hash matches the one against the retrieved user and return a custom error response? 

I think a lot of people would be grateful if you’d reconsider this change. Thanks again for all the hard work you guys put in to make it easier for us to create great apps 🙂

PR #202, opened 2021-01-19
[1.x] Allow vendor:publish routes

jrd-lewis:
This will allow users to publish the Fortify routes via the ```vendor:publish``` artisan command.


taylorotwell:
The installation instructions would publish these by default and I don't think they actually need to be published by default for most applications.

jrd-lewis:
As shown in https://laracasts.com/series/laravel-authentication-options/episodes/4, you actually have to dig to customize Fortify's routes while Jetstream publishes its routes when using ```vendor:publish```. This was to bring feature parity between the two packages.

Issue #209, opened 2021-01-26
Exception if password input is not a string


bonroyage:
- Fortify Version: 1.7.5
- Laravel Version: 8.24.0
- PHP Version: 7.4.1

### Description:
When providing a non-scalar input for the password field (e.g. an array), the Password rule will throw an exception for `mb_*` and `preg_match` functions:
- mb_strlen() expects parameter 1 to be string, array given (without additional requirements)
- mb_strtolower() expects parameter 1 to be string, array given (with `requireUppercase`)
- preg_match() expects parameter 2 to be string, array given (with `requireNumeric` and `requireSpecialCharacter`)

### Steps To Reproduce:
Exception:
```php
Route::get('validate', function () {
    $input = ['password' => ['foo' => 'bar']];
    
    $validator = \Illuminate\Support\Facades\Validator::make($input, [
        'password' => ['required', 'string', (new \Laravel\Fortify\Rules\Password())->requireSpecialCharacter()],
    ]);

    $validator->passes();

    return $validator->failed();
});
```

### Possible solutions:
Option 1: add `bail` before the Password rule

Option 2: check if the value is a string in the Password rule class (it's a direct copy of the `string` rule). I'd be happy to put this in a PR if you'd like to go this route.

```php
public function passes($attribute, $value)
{
    if(! is_string($value)) { // add this condition
        return false;
    }
    
    if ($this->requireUppercase && Str::lower($value) === $value) {
        return false;
    }

    if ($this->requireNumeric && ! preg_match('/[0-9]/', $value)) {
        return false;
    }

    if ($this->requireSpecialCharacter && ! preg_match('/[\W_]/', $value)) {
        return false;
    }

    return Str::length($value) >= $this->length;
}
```

driesvints:
Thanks. Sent in a PR here: https://github.com/laravel/fortify/pull/211

PR #211, opened 2021-01-28
[1.x] Fix password rule

driesvints:
This fixes a bug when a non-string value is password as input for a password. For example, as an array through an API request from a third party.

https://github.com/laravel/fortify/issues/209

### DIFF
--- a/src/Rules/Password.php
+++ b/src/Rules/Password.php
@@ -51,6 +51,8 @@ class Password implements Rule
      */
     public function passes($attribute, $value)
     {
+        $value = is_array($value) ? '' : (string) $value;
+
         if ($this->requireUppercase && Str::lower($value) === $value) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


src/Rules/Password.php
```diff
@@ -51,6 +51,8 @@ class Password implements Rule
      */
     public function passes($attribute, $value)
     {
+        $value = is_array($value) ? '' : (string) $value;
```
bonroyage, line 54: Would it make more sense to flip this to an `is_scalar($value) ? (string) $value : ''` to also capture any possible scenario where the input is an object, although that wouldn't really happen with validating requests but could happen if someone were to run the Validator manually.

The current approach would also convert any class that has __toString implemented, not sure if you intended it like that?

driesvints: Good suggestion, done :thu

PR #221, opened 2021-02-14
[1.x] The e-mail filter was added

Rene-Roscher:
The e-mail filter was added, otherwise mail addresses can be added in any way (no valid addresses)


driesvints:
The current email filter uses the default behavior and allows all valid email addresses. I'm not sure about what invalid email addresses you're talking about?

You can modify this in your own app if you want.

PR #226, opened 2021-02-24
Add environment variables to fortify config to allow overloading in env files

anubisthejackle:
This should make solving #77 simple, as all you need to do to change the default route is update your `.env` file. This is what I've done locally.

I also suggest this as it sticks to what we see in other config files that we work with on a daily basis, which allows us to have a standard location to make customizations to the framework, without modifying files, or requiring us to build entire classes to handle something as simple as a changed URL.

Which brings up the caveat to this: Anything beyond a basic URL change isn't handled this way, but I think that works fine. If you need more, there are other methods already available to account for that.

### DIFF
--- a/stubs/fortify.php
+++ b/stubs/fortify.php
@@ -16,7 +16,7 @@ return [
     |
     */
 
-    'guard' => 'web',
+    'guard' => env('FORTIFY_GUARD', 'web'),
 
     /*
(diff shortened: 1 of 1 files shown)


taylorotwell:
This file is exported to your own application so you can simply modify it according to your liking then.

PR #229, opened 2021-03-02
Use closure to add columns after an existing column

pavinthan:
To avoid  `ArgumentCountError`

```
  Too few arguments to function Illuminate\Database\Schema\Blueprint::after(), 1 passed in /var/www/html/database/migrations/2014_10_12_200000_add_two_factor_columns_to_users_table.php on line 18 and exactly 2 expected
```

Laravel Framework 8.30.0

Ref https://github.com/laravel/framework/blob/81ef9850cc388f2f92b868fb35ffb76f0c9a0f46/src/Illuminate/Database/Schema/Blueprint.php#L1543


taylorotwell:
Can't recreate such an error.

PR #234, opened 2021-03-08
[1.x] Fix two-factor forgotten login id from session

amiranagram:
In [this PR](https://github.com/laravel/fortify/pull/233) @driesvints added the enhancement I requested for two factor login page to redirect to login page if user hasn't previously submitted his username and password. 

Because the newly added method `hasChallengedUser` calls `challengedUser` method which pulls `login.id` from session the user will never be able to login because there is no `login.id` on the next (POST) request.


driesvints:
Thanks for catching this! I've sent in an alternative fix here: https://github.com/laravel/fortify/pull/235

amiranagram:
@driesvints Glad I could help.

Issue #236, opened 2021-03-10
The error for 2FA is returned with key "email"


mokhosh:
- Fortify Version: 1.7.7

### Description:
When I input a wrong 2FA code and get redirected back to the login page, the error message is returned under the key "email" which doesn't make any sense.

I have set `fortify.username` config to `login` and it makes more sense to at least use the `username` set in the config, if we are gonna send it under anything other than `two-factor-authentication-code`.

### Steps To Reproduce:
- input wrong 2FA code on challenge page
- `dd` the `$errors` on login page

driesvints:
Indeed. That seems odd to me as well. I'll check with @taylorotwell.

driesvints:
So we won't be able to change this atm because it'd break all Jetstream apps. The earliest we can change this is in the next Fortify version. You're free to send a PR to master if you like. We'll also need to update Jetstream's master branch in that case and bump the minimum version of Fortify to `^2.0` for its composer dependencies.

PR #238, opened 2021-03-12
[2.x] Use code instead of email for the key

mokhosh:
This should resolve https://github.com/laravel/fortify/issues/236

Also I was surprised to see that all the tests passed before and after the change. I would love to add tests for this as well, but I'm not sure how to test a two step verification challenge.
It shouldn't be an issue though, since there were no tests in the first place.

### DIFF
--- a/src/Http/Responses/FailedTwoFactorLoginResponse.php
+++ b/src/Http/Responses/FailedTwoFactorLoginResponse.php
@@ -23,6 +23,6 @@ class FailedTwoFactorLoginResponse implements FailedTwoFactorLoginResponseContra
             ]);
         }
 
-        return redirect()->route('login')->withErrors(['email' => $message]);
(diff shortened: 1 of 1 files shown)


driesvints:
@taylorotwell I checked Jetstream and I think this change can maybe still be made in 1.x as it's never directly used in Jetstream. Instead, errors are looped over here: https://github.com/laravel/jetstream/blob/2.x/resources/views/components/validation-errors.blade.php

So maybe we should target 1.x?

mokhosh:
@driesvints I guess it's ok for Jetstream, but this can be breaking for those who use Fortify and have customised 2FA flow.

I'm using `email` key in one of my websites just because it's the only way to know if there's a 2FA error now.

driesvints:
@mokhosh that's true.

taylorotwell:
Could be a breaking change.

PR #246, opened 2021-03-30
Always login same way

fbrissi:
Currently when customizing the login with `Fortify::authenticateUsing` when creating the user he logs in ignoring this customization this feature is for logging in when creating a user using the same rule as in login.

### DIFF
--- a/src/Http/Controllers/RegisteredUserController.php
+++ b/src/Http/Controllers/RegisteredUserController.php
@@ -4,6 +4,7 @@ namespace Laravel\Fortify\Http\Controllers;
 
 use Illuminate\Auth\Events\Registered;
 use Illuminate\Contracts\Auth\StatefulGuard;
+use Illuminate\Contracts\Support\Responsable;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Controller;
 use Laravel\Fortify\Contracts\CreatesNewUsers;
@@ -12,6 +13,8 @@ use Laravel\Fortify\Contracts\RegisterViewResponse;
 
 class RegisteredUserController extends Controller
 {
+    use AuthenticationPipeline;
+
     /**
      * The guard implementation.
      *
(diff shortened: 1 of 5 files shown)


src/Http/Controllers/RegisteredUserController.php
```diff
@@ -46,15 +48,15 @@ public function create(Request $request): RegisterViewResponse
      *
      * @param  \Illuminate\Http\Request  $request
      * @param  \Laravel\Fortify\Contracts\CreatesNewUsers  $creator
-     * @return \Laravel\Fortify\Contracts\RegisterResponse
+     * @return \Laravel\Fortify\Contracts\RegisterResponse | \Illuminate\Http\RedirectResponse
      */
     public function store(Request $request,
-                          CreatesNewUsers $creator): RegisterResponse
+                          CreatesNewUsers $creator)
```
driesvints, line 54: This is a breaking change to the method signature.

fbrissi: I know, but when, try login and failed, weee have an `\Illuminate\Http\RedirectResponse` and `Laravel\Fortify\Contracts\RegisterResponse` extend `Illuminate\Contracts\Support\Responsable`, we don't have no return in the same hierarchy to actually use polymorphism.

driesvints: Anyone only expecting a `RegisterResponse` will see their apps fail if something else is returned. This should be sent to master.

fbrissi: Then, login only with the same rules, and always return `RegisterResponse`?

```PHP
    public function store(Request $request,
                          CreatesNewUsers $creator): RegisterResponse
    {
        event(new Registered($user = $creator->create($request->all())));

        $this->loginPipeline($request)->then(function () {
        });

        return app(RegisterResponse::class);
```

fbrissi: This looks good?

taylorotwell:
Don't think I will change this for now. You should still be able to listen to the Login event if you need to do something in response to authentication.

Issue #253, opened 2021-04-19
Release 1.7.9 has broken all my websites


mokhosh:
So I've been getting support tickets that my password reset forms aren't working, and I tracked it down, and seems like this PR https://github.com/laravel/fortify/pull/245 by dear @driesvints has a really bad breaking change.

I never use password confirmation in my applications, I use "show/hide" button instead for the user to be sure of what they've entered if they want to.

@driesvints mentions in the PR he doesn't know why this hasn't been done before, I think this hasn't been done because we already validate the password in a customisable way in whatever class we pass to `ResetPasswordsUsing` closure. Right?

mokhosh:
btw i don't want to sound demanding with this title, I just wanted to acknowledge that this was a serious breaking change, not just a patch and it should have at least been reflected in semver.

driesvints:
This has nothing to do with password confirmation. The code that changed is for resetting passwords. How would this affect you?

Also, please don't ignore the issue template next time.

mokhosh:
This makes it required to confirm your password when you want to reset it.

mokhosh:
- Fortify Version: 1.7.10
- Laravel Version: 8.35.1
- PHP Version: 8.0.3
- Database Driver & Version: mysql 8.0.23

### Description:
I think the description above was enough, and the versions were irrelevant as I mentioned the exact PR where the regression has started.

### Steps To Reproduce:
Update fortify to 1.7.9 and above, or worse if your using jetstream just update composer and it will update fortify for you and you'll have no easy way of reverting back to an older version.

driesvints:
Thanks, I've sent in a PR for this: https://github.com/laravel/fortify/pull/254

You can btw always require Fortify explicitly in your composer.json.

> I think the description above was enough, and the versions were irrelevant as I mentioned the exact PR where the regression has started.

Please don't ever ignore an issue template. It helps us with identifying potential bugs more quickly.

mokhosh:
> Please don't ever ignore an issue template. It helps us with identifying potential bugs more quickly.

Ok, will keep in mind.

Thanks.

PR #254, opened 2021-04-19
[1.x] Remove password confirmation requirement

driesvints:
In some situations you don't need the password confirmation if this isn't part of the password reset flow.

See https://github.com/laravel/fortify/issues/253


taylorotwell:
What if some people do want confirmation? I feel like it's more common to see sites have a confirm password field than not?

PR #259, opened 2021-04-28
create custom redirect to

akunbeben:
This PR produces `Fortify::redirectAfterLoginUsing($callback)` and `Fortify::redirectAfterRegisterUsing($callback)` callbacks, the idea is based on ( #77 ). it gives developer **more options to customize** their redirect using callback or create another implementation of `Fortify\Http\Responses`

### DIFF
--- a/src/Fortify.php
+++ b/src/Fortify.php
@@ -4,7 +4,9 @@ namespace Laravel\Fortify;
 
 use Laravel\Fortify\Contracts\ConfirmPasswordViewResponse;
 use Laravel\Fortify\Contracts\CreatesNewUsers;
+use Laravel\Fortify\Contracts\LoginResponse;
 use Laravel\Fortify\Contracts\LoginViewResponse;
+use Laravel\Fortify\Contracts\RegisterResponse;
 use Laravel\Fortify\Contracts\RegisterViewResponse;
 use Laravel\Fortify\Contracts\RequestPasswordResetLinkViewResponse;
 use Laravel\Fortify\Contracts\ResetPasswordViewResponse;
@@ -237,6 +239,17 @@ class Fortify
     public static function createUsersUsing(string $callback)
     {
         return app()->singleton(CreatesNewUsers::class, $callback);
+    }
+
+    /**
+     * Register a callback that should be used to redirect user after login.
+     *
+     * @param  callable  $callback
+     * @return void
+     */
+    public static function redirectAfterLoginUsing(callable $callback)
+    {
+        return app()->singleton(LoginResponse::class, $callback);
     }
 
     /**
--- a/src/Fortify.php
+++ b/src/Fortify.php
@@ -253,6 +253,17 @@ class Fortify
     }
 
     /**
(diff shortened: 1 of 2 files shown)


akunbeben:
Then, the implementation to the project would be like this.

```php
<?php

namespace App\Providers;

use App\Actions\Fortify\CreateNewUser;
use App\Actions\Fortify\ResetUserPassword;
use App\Actions\Fortify\UpdateUserPassword;
use App\Actions\Fortify\UpdateUserProfileInformation;
use Illuminate\Cache\RateLimiting\Limit;
use Illuminate\Http\Request;
use Illuminate\Support\Facades\RateLimiter;
use Illuminate\Support\ServiceProvider;
use Laravel\Fortify\Contracts\LoginResponse;
use Laravel\Fortify\Fortify;

class FortifyServiceProvider extends ServiceProvider
{
    /**
     * Register any application services.
     *
     * @return void
     */
    public function register()
    {
        //
    }

    /**
     * Bootstrap any application services.
     *
     * @return void
     */
    public function boot()
    {
        Fortify::createUsersUsing(CreateNewUser::class);
        Fortify::updateUserProfileInformationUsing(UpdateUserProfileInformation::class);
        Fortify::updateUserPasswordsUsing(UpdateUserPassword::class);
        Fortify::resetUserPasswordsUsing(ResetUserPassword::class);

        Fortify::redirectAfterLoginUsing(function () {
            return request()->wantsJson()
                ? response('', 201)
                : redirect()->intended('custom-url-after-login');
        });

        Fortify::redirectAfterRegisterUsing(function () {
            return request()->wantsJson()
                ? response('', 201)
                : redirect()->intended('custom-url-after-registered');
        });

        RateLimiter::for('login', function (Request $request) {
            return Limit::perMinute(5)->by($request->email . $request->ip());
        });

        RateLimiter::for('two-factor', function (Request $request) {
            return Limit::perMinute(5)->by($request->session()->get('login.id'));
        });
    }
}
```

taylorotwell:

akunbeben:
Thanks for your response @taylorotwell .

I think this would provide options to customize their redirection. referring to issue (#77) that labeled with enhancement, this would be a solution to the issue.

The usage scenario would be like this.

Case 1:
If I need to customize some complex redirection. The solution is:
Create a new implementation of ResponseContracts. Then bind the new implementation in the `FortifyServiceProvider.php`

Case 2:
If I only need a simple redirect to somewhere else that I wanted to customize. The solution is:
So, I just need to call these callback method `Fortify::redirectAfterLoginUsing()` or `Fortify::redirectAfterRegisterUsing` in the `FortifyServiceProvider.php` like this example below:

```php
Fortify::redirectAfterLoginUsing(function () {
    return request()->wantsJson()
        ? response(['two_factor' => false], 200)
        : redirect()->intended('custom-url-after-login');
});

Fortify::redirectAfterRegisterUsing(function () {
    return request()->wantsJson()
        ? response('', 201)
        : redirect()->intended('custom-url-after-registered');
});
```

PR #264, opened 2021-05-17
Add redirect config after login, logout and registration

trbsi:
In this PR I added posibility to set redirection route after logout, login and registration. Registration route can be set inside config/fortify.php

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -10,6 +10,11 @@ return [
     'email' => 'email',
     'views' => true,
     'home' => '/home',
+    'redirect' => [
+        'login' => '/home',
+        'registration' => '/home',
+        'logout' => '/login',
+    ],
     'prefix' => '',
     'domain' => null,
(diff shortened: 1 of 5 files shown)


driesvints:
This is a breaking change because the current config relies on `fortify.home`. You're changing this back to the default which some people might have adjusted.

trbsi:
That's true, if people get this change they will have to rely on this part

```
    'redirect' => [
        'login' => RouteServiceProvider::HOME,
        'registration' => RouteServiceProvider::HOME,
        'logout' => '/login',
    ],
```

What would be your suggestion here?
I think LogoutResponse.php can be adjusted and set default redirection to "/"?

taylorotwell:
I dunno - maybe it is safer to just tell users to do this:

image

trbsi:
So basically this: https://github.com/laravel/docs/pull/7063 ?

PR #267, opened 2021-05-20
returns json response

ErwinLiemburg:
The VerifyEmailController only had support for returning a redirect to a page. But when using API to confirm a new user, a JSON reply is expected. Changed the code to support the functionality, that when required the response is in JSON format.

Had to remove the return type, to be able to support JSON reply.

If anything can be improved, please be my guest.


ErwinLiemburg:
By default the link in the email links to the 'default' laravel verification page. But when you use laravel as a backend API with a SPA, it seems weird that the link is sent to a page different from the SPA. By changing the link in the verification email, sending it to the SPA, the SPA can therefor send a verification request, and handle the result. The SPA would require a JSON reply, and not a redirect to another page.

Issue #201, opened 2021-01-17
Current 2FA solution can lock users out of their accounts


LeoniePhiline:
- Version: v1.7.4
- Laravel Version: v8.22.1
- PHP Version: v7.4.13

### Description:

The current two factor authentication solution sets 2fa to enabled without requiring a confirmation (via TOTP) that the authenticator app is actually set up.

### Steps To Reproduce:

The current solution works like this:

1. `POST /user/two-factor-authentication`, the user's `two_factor_secret` is stored. (2fa is now enabled!)
2. `GET /user/two-factor-qr-code`, show QR code and ask user to scan the code with their app.
3. `GET /user/two-factor-recovery-codes`, show recovery codes and ask the user to save them.
4. User abandons the process before setting up their authenticator app or saving the recovery codes and is now **locked out of their account**.

They could abandon the process because they first need to choose one of the many TOTP generators in the app store, and get side tracked, or their session times out, or they click the back button, or close their tab, or their computer crashes, ... Definitely 2fa must not be enabled before it is confirmed by a generated OTP.

### How To Fix It:

1. `GET /user/two-factor-qr-code` generates a QR code from a new two factor secret that is **stored in the session**.
2. `GET /user/two-factor-recovery-codes`, show the recovery codes to the user, ask them to save them. 
3. The user is asked to set up their authenticator app with this QR code and enter a resulting TOTP code. This confirms that they have set up their authenticator (else they cannot generate a valid code), and can also (by written explanation) be used as confirmation that recovery codes were stored in a safe place.
4. `POST /user/two-factor-authentication`, receives a new parameter: `code`. **The code is validated using the two factor secret stored in the session.**
5. If the code is valid, **the two factor secret from the session is written into the user table** (= enabling 2fa, now for real). If not, then the response indicates that the user did not enter a valid code and must re-try.

driesvints:
This issue basically ties into https://github.com/laravel/jetstream/issues/74. While I agree it would be better if we required it, it's opinionated for sure. We could make this an opt-in feature for Fortify 1.x so we can implement it in Jetstream together with https://github.com/laravel/jetstream/issues/74.

LeoniePhiline:
Having it as an opt in feature for now to avoid the breaking change until the next major version sounds good to me. 

I kindly disagree that this is opinionated, though. ;-)

——

By the way, is there a plan for a RequireSecondFactor confirmation middleware (with timeout) that works like the laravel core RequirePassword middleware?

[...]

PR #275, opened 2021-06-02
Add opt-in feature to confirm two factor setup

dmandrade:
As described in [#201](https://github.com/laravel/fortify/issues/201), the current 2FA solution can be activated without the user actually having done the proper configuration preventing them from logging in.

I added a new route and an action to confirm 2FA setup using a TOTP code.

I also moved the contents of the EnableTwoFactorAuthentication action to a new one called GenerateTwoFactorAuthenticationSecret but keep the original to avoid breaking changes.

### DIFF
--- a/database/migrations/2021_06_01_145800_add_two_factor_enabled_column_to_users_table.php
+++ b/database/migrations/2021_06_01_145800_add_two_factor_enabled_column_to_users_table.php
@@ -1,0 +1,34 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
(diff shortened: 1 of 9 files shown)


dmandrade:
> This most likely needs to go into master as there are quite a few breaking changes.

I fixed the problem and passed on now, there's no breaking changes anymore.
Can I make a new PR for 1.x or master?

driesvints:
@dmandrade think it's best to go into master regardless. The new columns will disable 2factor for all existing apps.

PR #276, opened 2021-06-03
[2.x] Confirm 2FA setup before enabled it

dmandrade:
As described in #201, the current 2FA solution can be activated without the user actually having done the proper configuration preventing them from logging in.

I added new tests, route and an action to confirm 2FA setup using a TOTP code.


taylorotwell:
I will revisit this myself at some point.

ossicoinc:
Why was this one closed / not merged?

joelbutcher:
@ossicoinc see Taylors comment - it's something the Laravel team plan on implementing themselves

Issue #277, opened 2021-06-06
Consider changing parameters on contracts


Velka-DEV:
- Fortify Version: 1.7
- Laravel Version: 8.40
- PHP Version: 7.3
- Database Driver & Version: MySQL

### Description:

Actually array's of data input is pass to contracts functions (Actions like CreateNewUser)

It's not really cool because we can't access other request data, like actually i need to retrieve a cookie on the register request but i can't access it because we don't have the request

Or maybe i miss something

driesvints:
If you want to propose something feel free to send in a PR 👍

Velka-DEV:
PR wating for review: https://github.com/laravel/fortify/pull/278

PR #280, opened 2021-06-08
Switched create() registering param from request inputs to raw request …

Velka-DEV:
Related to issue: #277


Velka-DEV:
@taylorotwell Can you explain more about why you refused the PR ?

PR #282, opened 2021-06-09
Use remember property value vs. presence to determine if the user is remembered 

aidan-casey:
When making a login request to Fortify, the user is still remembered when the `remember` property is set to a falsy value. This is because Fortify uses the `filled()` method to check for the remember property.

I think this can be confusing for users and when reading the [documentation](https://laravel.com/docs/8.x/fortify#authentication), it seems like a bug in the way it is worded.

> In addition, a boolean remember field may be provided to indicate that the user would like to use the "remember me" functionality provided by Laravel.

This PR just updates Fortify to utilize the `boolean()` method to check for the remember property just like Breeze does.

### DIFF
--- a/src/Actions/AttemptToAuthenticate.php
+++ b/src/Actions/AttemptToAuthenticate.php
@@ -52,7 +52,7 @@ class AttemptToAuthenticate
 
         if ($this->guard->attempt(
             $request->only(Fortify::username(), 'password'),
-            $request->filled('remember'))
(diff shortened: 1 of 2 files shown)


taylorotwell:
That could be a breaking change for existing applications.

aidan-casey:
@taylorotwell Totally get it. Would it be helpful for me to PR to master?

PR #283, opened 2021-06-15
Limit Uppercase, Numeric, Special Character via a parameter.

devbapidey:
Previously, Only one Uppercase, Numeric, Special Character is accepted. Now we can define how much character we need via parameter.


driesvints:
Seems like something went wrong with branching here. Please try again.

devbapidey:
Sorry I didn't understand. Did my code fail? You mentioned branch, please explain.

driesvints:
There's commits from 1.x on this PR. Please branch anew from master and resubmit.

Issue #272, opened 2021-05-27
add support length for requireUppercase requireNumeric requireSpecialCharacter for Password class


devbapidey:
We need a minimum count for uppercase, numeric, special character. So before
```
(new Password)
                ->length(12)
                ->requireUppercase()
                ->requireNumeric()
                ->requireSpecialCharacter()
```
after
`(new Password)
                ->length(12)
                ->requireUppercase(3)
                ->requireNumeric(3)
                ->requireSpecialCharacter(3)`

driesvints:
Heya, thanks for submitting this.

This seems like a feature request or an improvement so I'm moving this to the ideas repository instead. It's best to post these in the ideas repository in the future to get support for your idea. After that you may send a PR to the framework. Please only use the `laravel/framework` issue tracker to report bugs and issues with the framework.

Thanks!

driesvints:
Sorry, realized this is the fortify repo.

Feel free to attempt a PR if you want. This probably needs to go into master.

PR #284, opened 2021-06-15
[2.x] Limit Uppercase, Numeric, Special Character via a parameter

devbapidey:
Previously, Only one Uppercase, Numeric, Special Character is accepted. Now we can define how much character we need via parameter.


devbapidey:
[I requested this feature on that issue.](https://github.com/laravel/fortify/issues/272)

taylorotwell:

PR #286, opened 2021-06-16
Requirements of UC, Numeric, SC not limited to singular. Now it can define via an int argument.

devbapidey:
`requireUppercase`, `requireNumeric`, `requireSpecialCharacter` function now receive a int argument. That argument helps dev to defines how much requirement. By default, it is '0' if func not called. If called, it's '1'.

Previously I PR it. And rejected with this comment - "To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include.".
I totally agree with it. But that time my code was wrong. So I retry. If everything goes fine, maybe in the future this code will merge with master.

### DIFF
--- a/src/Rules/Password.php
+++ b/src/Rules/Password.php
@@ -15,21 +15,21 @@ class Password implements Rule
     protected $length = 8;
 
     /**
-     * Indicates if the password must contain one uppercase character.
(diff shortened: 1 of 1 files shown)


taylorotwell:

PR #293, opened 2021-07-08
[1.x] Add config to allow signed:relative in email verification

dseidl:
Hello 👋 

this PR enables the option to use the signed:relative middleware as described in [33137](https://github.com/laravel/framework/pull/33137). It defaults to false.

I used fortify with a separate frontend, and changed the email verification link through `createUrlUsing` setting the absolute parameter to false. This allowed me to prefix my frontend domain and send the verify request from there. For that to go through it is required to also set the `signed:relative` middleware.

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -9,6 +9,7 @@ return [
     'username' => 'email',
     'email' => 'email',
     'views' => true,
+    'signed_relative' => false,
     'home' => '/home',
     'prefix' => '',
     'domain' => null,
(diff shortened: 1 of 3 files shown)


driesvints:
In this case I'd simply overwrite the route itself in your own `web.php` routes file.

taylorotwell:
@driesvints cna you actually overwrite the route in that way... I think the FortifyServiceProvider would be booted first?

driesvints:
@taylorotwell yes, just tried that out. It's the same thing which we suggest to do when overwriting the webhook route in, for example, Cashier Stripe.

taylorotwell:
OK - then @dseidl I would suggest just defining the route how you like in your own routes file as @driesvints suggests.

PR #300, opened 2021-08-02
[1.x] Prevent duplicate query using user from request

### LINKED ISSUE #120: AttemptToAuthenticate is not throwing validation exceptions by default
- Fortify Version: 1.6.1
- Laravel Version: 8.9.0
- PHP Version: 7.4.9
- Database Driver & Version: MySQL 5.7.31

### Description:

This is the first time I'm using Fortify. I was trying to customize the login flow, because I want to change the default error bag on the `ValidationException`. But I noticed that the `throwFailedAuthenticationException` method on the `AttemptToAuthenticate` action is not being called, because the `RedirectIfTwoFactorAuthenticatable` action is calling `validateCredentials`, despite that I'm not using 2FA.

So the login pipeline is stopping at `RedirectIfTwoFactorAuthenticatable::validateCredentials()`, and the `AttemptToAuthenticate`, does not do anything, even though it's supposed to attempt to login or throw a validation exception.

### Steps To Reproduce:
1. Override `AttemptToAuthenticate`:
```php
use Illuminate\Validation\ValidationException;
use Laravel\Fortify\Actions\AttemptToAuthenticate;

class CustomAttemptToAuthenticate extends AttemptToAuthenticate 
{
    protected function throwFailedAuthenticationException($request)
    {
        try {
            parent::throwFailedAuthenticationException($request);
        } catch (ValidationException $e) {
            throw $e->errorBag('login');
        }
    }
}
```
2. Override login pipeline:
```php

Fortify::loginThrough(function () {
    return [
        config('fortify.limiters.login') ? null : EnsureLoginIsNotThrottled::class,
        RedirectIfTwoFactorAuthenticatable::class,
        CustomAttemptToAuthenticate::class,
        PrepareAuthenticatedSession::class,
    ];
});
```

3. Test the custom error bag:
```php
$response = $this->post('login', [
    'email' => 'invalid@email.com',
    'password' => 'password',
]);

 // this assertion fails
$response->assertSessionHasErrors('email', null, 'login');
```

4. Note the exception stack trace:
```php
$this->withoutExceptionHandling();

1   vendor/laravel/fortify/src/Actions/RedirectIfTwoFactorAuthenticatable.php:98
    Illuminate\Validation\ValidationException::withMessages()

2   vendor/laravel/fortify/src/Actions/RedirectIfTwoFactorAuthenticatable.php:80
    Laravel\Fortify\Actions\RedirectIfTwoFactorAuthenticatable::throwFailedAuthenticationException()

3   vendor/laravel/framework/src/Illuminate/Support/helpers.php:263
    Laravel\Fortify\Actions\RedirectIfTwoFactorAuthenticatable::Laravel\Fortify\Actions\{closure}()

4   vendor/laravel/fortify/src/Actions/RedirectIfTwoFactorAuthenticatable.php:82
    tap()

5   vendor/laravel/fortify/src/Actions/RedirectIfTwoFactorAuthenticatable.php:50
    Laravel\Fortify\Actions\RedirectIfTwoFactorAuthenticatable::validateCredentials()

6   vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:167
    Laravel\Fortify\Actions\RedirectIfTwoFactorAuthenticatable::handle()

7   vendor/laravel/fortify/src/Actions/EnsureLoginIsNotThrottled.php:39
    Illuminate\Pipeline\Pipeline::Illuminate\Pipeline\{closure}()

8   vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:167
    Laravel\Fortify\Actions\EnsureLoginIsNotThrottled::handle()

9   vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:103
    Illuminate\Pipeline\Pipeline::Illuminate\Pipeline\{closure}()

10  vendor/laravel/fortify/src/Http/Controllers/AuthenticatedSessionController.php:60
    Illuminate\Pipeline\Pipeline::then()


```

ricardogobbosouza:
Resolve #120

@driesvints Both classes have very similar parts, do you think it's valid to abstract to another class? Ex: `AbstractAuthenticate`

### DIFF
--- a/src/Actions/AttemptToAuthenticate.php
+++ b/src/Actions/AttemptToAuthenticate.php
@@ -72,7 +72,7 @@ class AttemptToAuthenticate
         $user = call_user_func(Fortify::$authenticateUsingCallback, $request);
 
         if (! $user) {
-            $this->fireFailedEvent($request, $user);
+            $this->fireFailedEvent($request);
 
             return $this->throwFailedAuthenticationException($request);
         }
@@ -103,12 +103,11 @@ class AttemptToAuthenticate
      * Fire the failed authentication attempt event with the given arguments.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @param  \Illuminate\Contracts\Auth\Authenticatable|null  $user
      * @return void
      */
-    protected function fireFailedEvent($request, $user = null)
+    protected function fireFailedEvent($request)
     {
-        event(new Failed(config('fortify.guard'), $user, [
+        event(new Failed(config('fortify.guard'), null, [
             Fortify::username() => $request->{Fortify::username()},
             'password' => $request->password,
         ]));
--- a/src/Actions/AttemptToAuthenticate.php
+++ b/src/Actions/AttemptToAuthenticate.php
@@ -50,7 +50,7 @@ class AttemptToAuthenticate
             return $this->handleUsingCustomCallback($request, $next);
         }
 
-        if ($this->guard->attempt(
+        if ($request->user() || $this->guard->attempt(
             $request->only(Fortify::username(), 'password'),
             $request->filled('remember'))
         ) {
@@ -72,7 +72,7 @@ class AttemptToAuthenticate
         $user = call_user_func(Fortify::$authenticateUsingCallback, $request);
 
         if (! $user) {
-            $this->fireFailedEvent($request);
+            $this->fireFailedEvent($request, $user);
 
             return $this->throwFailedAuthenticationException($request);
         }
@@ -103,11 +103,12 @@ class AttemptToAuthenticate
      * Fire the failed authentication attempt event with the given arguments.
      *
      * @param  \Illuminate\Http\Request  $request
+     * @param  \Illuminate\Contracts\Auth\Authenticatable|null  $user
      * @return void
      */
-    protected function fireFailedEvent($request)
+    protected function fireFailedEvent($request, $user = null)
     {
-        event(new Failed(config('fortify.guard'), null, [
+        event(new Failed(config('fortify.guard'), $user, [
             Fortify::username() => $request->{Fortify::username()},
             'password' => $request->password,
         ]));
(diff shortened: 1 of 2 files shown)


src/Actions/AttemptToAuthenticate.php
```diff
@@ -103,11 +103,12 @@ protected function throwFailedAuthenticationException($request)
      * Fire the failed authentication attempt event with the given arguments.
      *
      * @param  \Illuminate\Http\Request  $request
+     * @param  \Illuminate\Contracts\Auth\Authenticatable|null  $user
      * @return void
      */
-    protected function fireFailedEvent($request)
+    protected function fireFailedEvent($request, $user = null)
```
driesvints, line 109: I'm not sure why you made this change? On both of the places where you call and pass the `$user` you know that they're not set (see the ternary check in the if statement before calling the method). The user will always be `null`.

driesvints:
>  Both classes have very similar parts, do you think it's valid to abstract to another class? 

@ricardogobbosouza yes that sounds good but maybe in a different PR so we can keep the focus at what this fixes.

src/Actions/AttemptToAuthenticate.php
ricardogobbosouza: ok, removed!

ricardogobbosouza:
> > Both classes have very similar parts, do you think it's valid to abstract to another class?
> 
> @ricardogobbosouza yes that sounds good but maybe in a different PR so we can keep the focus at what this fixes.

@driesvints  I'll do it

taylorotwell:
This has come up before but I don't want to login anyone in if they haven't passed the two factor challenge - which this change does.

driesvints:
@ricardogobbosouza can you send in a separate PR for just the `$request->user()` check?

ricardogobbosouza:
@driesvints Unfortunately, just `$request->user()` won't solve the problem, as the user hasn't been defined yet.

But we can't login without confirming the two-factor authentication as @taylorotwell  said.

Maybe define the user in the request here 

https://github.com/laravel/fortify/blob/1.x/src/Actions/RedirectIfTwoFactorAuthenticatable.php#L56
```php
$request['challengedUser'] = $user
```

and here https://github.com/laravel/fortify/blob/1.x/src/Actions/AttemptToAuthenticate.php#L52
check if this user exists and log in, so this second query is not executed.
```php
if ($user = $request['challengedUser']) {
    $this->guard->login ($user, $request->full('remember'));
}
```

driesvints:
@ricardogobbosouza can you reply that to the issue so we can keep track of it? Otherwise I'm gonna lose sight of it. Thanks

ricardogobbosouza:
@driesvints this should solve https://github.com/laravel/fortify/pull/304 😄

PR #307, opened 2021-08-24
Feature/reset password redirect

rcerljenko:
Hi,

Improved PR with default fallback to login route as @taylorotwell mentioned.

Cheers!

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -21,6 +21,7 @@ return [
         'password-confirmation' => null,
         'register' => null,
         'email-verification' => null,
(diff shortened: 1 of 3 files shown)


dmanva:
Sorry if I misunderstood something, but shouldn't 'redirects' array also be in [publishable config file](https://github.com/laravel/fortify/blob/1.x/stubs/fortify.php) ?

rcerljenko:
@dmanva you right it should
they probably forgot to add it in previous release (1.8.0) where redirects were introduced

Issue #308, opened 2021-08-28
Email Verification for API users


rcerljenko:
- Fortify Version: 1.8.1
- Laravel Version: 8.57
- PHP Version: 8.0.10
- Database Driver & Version: N/A

### Description:

This is also [Laravel related](https://laravel.com/docs/8.x/verification#the-email-verification-handler) but since I'm using Fortify to handle my auth I'm posting it here...

Email Verification controller assumes that user has an active session while trying to mark email as verified => [here](https://github.com/laravel/fortify/blob/1.x/src/Http/Requests/VerifyEmailRequest.php). Also verification routes use `auth` middleware on them => [here](https://github.com/laravel/fortify/blob/1.x/routes/routes.php#L85).

That's great for default auth session guard (web guard) but API users use different (stateless) guards such as Sanctum or JWT so they don't have an active session and thus you can't verify your email once you click on a link that has been sent to your email account.

I'm thinking that maybe we loose the `auth` middleware on that routes and fetch the user from database by decrypted ID from the signed route itself.

If interested in this I can provide a PR with the proposed solution.

driesvints:
Why can't send you a verification request through the api with an active token? The user should already be created with a valid password and email address.

rcerljenko:
@driesvints but registered user receives the verification link on email with laravel default flow. We can maybe override that link creation and include JWT token on that link so that is sent in the verification email?

driesvints:
@rcerljenko good point. In that case I'd not do that since that could lead to abuse. Probably retrieving the user from the database by its ID like you proposed is a good solution. Can you send in a PR? Thanks.

rcerljenko:
@driesvints yes I can  :)

PR #309, opened 2021-08-28
Feature/stateless email verification

rcerljenko:
Implemented stateless email verification but kept the session one just in case.

It fixes https://github.com/laravel/fortify/issues/308.


rcerljenko:
@driesvints tests should probably be adjusted for new feature... any idea how to do this?

driesvints:
First of all, there's tons of code style changes here that are a bit of noise and make the PR hard to review. Please revert these. Also, these changes are very much breaking and should be sent to master. Please open an issue to master without code style changes and let's take it from there.

rcerljenko:
@driesvints got it.. that's my code editor making those style changes.. it should be pretty simple without those..

PR #310, opened 2021-08-28
Feature/stateless email verification

rcerljenko:
Re-targeted for master branch.


taylorotwell:

rcerljenko:
@taylorotwell @driesvints could I ask why was it declined? We covered the topic here => https://github.com/laravel/fortify/issues/308.
I know that tests are failing but besides that was anything else wrong?

driesvints:
@rcerljenko I've pinged Taylor about this

rcerljenko:
@driesvints thanks!

Also, a little explanation about why we need an extra config for verification model.

We could avoid this by putting the User model directly as a param of the controller so that is route model binded but I wanted to be configurable since somebody could have different auth model for this.
Everything else should be straightforward...

Ping me if you need any aditional info.

driesvints:
@rcerljenko we've backlogged in our internal Trello. So for right now we're not taking any action on this, sorry.

rcerljenko:
@driesvints ok... thanks for the update

Danoctum:
Will this ever be accepted?

Fortify is not a true headless authentication library, because it is very much reliant on a specific way of implementing a front-end to verify an email address. If I'd like to use Sanctum for a mobile app, and Fortify for authentication, I'd always need a website where a user can login in order for the email to be verified.

PR #313, opened 2021-09-06
allow verification rate limiter to be configurable

bastinald:
this will allow us to implement our own inertia responses for verification route throttling.

right now it is hard-coded so it's impossible to change.


bastinald:
Thanks for merging. Any idea when this will be released?

driesvints:
@bastinald tomorrow

Issue #314, opened 2021-09-19
Add route suffix name into fortify routes


ousid:
- Laravel: `v8.45`
- Fortify: `v1.6.1`
- PHP:  `8.0.10`

### Description
When I'm working on a project using Jetstream and fortify for authentication scaffolding, I faced a situation to ignore the default routes and create custom routes just for adding suffix for the `route names`, I think it would be a better idea to add `as` option on the route configuration and set it to null as the `domain` option.

### Steps to Reproduce
Sometimes we need to check if the route belongs to the `authentication routes` or a normal route to make certain operations (such as hide the `navbar` or something). So, instead of let's say

```blade
@if(! request()->routeIs('login', 'register', '...')
   Show navbar
@endif
```

We can just do this

```blade
@if(! request()->routeIs('auth.*'))
   show navbar
@endif
```

### The Solution
The solution actually is pretty simple, we need to add the `as` option into `configureRoutes` inside `fortifyServiceProvider` like this:

```diff
    protected function configureRoutes()
    {
        if (Fortify::$registersRoutes) {
            Route::group([
                'namespace' => 'Laravel\Fortify\Http\Controllers',
+                'as'  => config('fortify.route_suffix', null),
                'domain' => config('fortify.domain', null),
                'prefix' => config('fortify.prefix'),
            ], function () {
                $this->loadRoutesFrom(__DIR__.'/../routes/routes.php');
            });
        }
    }
```

and adding into `stubs/fortify.php` the following:

```diff
    'prefix' => '',
    'domain' => 'auth.' . config('app.url'),
+   'route_suffix'  => null,
```

I hope this make sense, and if it is, I can make a PR for it.

driesvints:
Seems like your PR was rejected, sorry.

PR #316, opened 2021-09-22
add `as` route paramter

ousid:
This PR is followed issue https://github.com/laravel/fortify/issues/314, I created before. And It contains the following:

### Description
When I'm working on a project using Jetstream and fortify for authentication scaffolding, I faced a situation to ignore the default routes and create custom routes just for adding suffix for the `route names`, I think it would be a better idea to add `as` option on the route configuration and set it to null as the `domain` option.

### Steps to Reproduce
Sometimes we need to check if the route belongs to the `authentication routes` or a normal route to make certain operations (such as hide the `navbar` or something). So, instead of let's say

```blade
@if(! request()->routeIs('login', 'register', '...')
   Show navbar
@endif
```

We can just do this

```blade
@if(! request()->routeIs('auth.*'))
   show navbar
@endif
```

### The Solution
The solution actually is pretty simple, we need to add the `as` option into `configureRoutes` inside `fortifyServiceProvider` & `fortify` config file.

### DIFF
--- a/src/FortifyServiceProvider.php
+++ b/src/FortifyServiceProvider.php
@@ -129,6 +129,7 @@ class FortifyServiceProvider extends ServiceProvider
                 'namespace' => 'Laravel\Fortify\Http\Controllers',
                 'domain' => config('fortify.domain', null),
                 'prefix' => config('fortify.prefix'),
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #317, opened 2021-10-04
add name for login route

smortexa:


taylorotwell:
The GET version is already named and the URIs are the same.

PR #318, opened 2021-10-10
Use provider methods instead of model for 2FA

makstech:
### Description:
Current 2FA validation ignores custom authentication providers. When there's a provider defined that extends `Illuminate\Auth\EloquentUserProvider` with a custom query in `newModelQuery` method, it is ignored and instead, user search is done by calling on models, so it bypasses  `newModelQuery` method.

### Steps To Reproduce:
1. Create a provider that extends `Illuminate\Auth\EloquentUserProvider`.
2. Overwrite `newModelQuery` method and add to query something that would certainly not return a user, like 
```php
    protected function newModelQuery($model = null)
    {
        return parent::newModelQuery($model)
            ->where('id', '=', -1);
    }
```
3. Try to log in with a gate that has the previously created provider.
4. `\Laravel\Fortify\Actions\RedirectIfTwoFactorAuthenticatable::validateCredentials` will still find the user and will try to validate credentials.

### DIFF
--- a/src/Actions/RedirectIfTwoFactorAuthenticatable.php
+++ b/src/Actions/RedirectIfTwoFactorAuthenticatable.php
@@ -73,9 +73,11 @@ class RedirectIfTwoFactorAuthenticatable
             });
         }
 
-        $model = $this->guard->getProvider()->getModel();
+        $user = $this->guard->getProvider()->retrieveByCredentials([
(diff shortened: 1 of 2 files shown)


mpyw:
@driesvints I guessed the real reverted reason here: https://github.com/laravel/fortify/pull/189#discussion_r906661504
Actually this PR #318 does not have any problems, LGTM.

PR #325, opened 2021-11-04
2FA Confirm & More

WadeShuler:
### What this does

 - Adds two factor confirmation to protect pages with two factor, for improved security.
 - Redirects back to two factor page on invalid code instead of the login page (super annoying).
 - Adds more events so we can better inform the user of what is going on with their account.
 - Allows the login process to count as confirmation, so they don't get immediately asked again (optional).
 - Prevents the password reset page from exposing valid email addresses (valid code check).
 - Allows changing the number of recovery codes generated. Default is still 8.
 - Organizes the files in the stubs directory.

This does not break BC, though a few extra things need to be added to utilize some of the new features.

### Fortify helper class

In order to add two factor confirmation, there are a few helper methods:

 - `Fortify::confirmTwoFactorView()` - The view file for the two factor confirmation form (required)
 - `Fortify::userTwoFactorNotEnabledView()` - Only needed if route is ONLY 2FA protected and user has it disabled
 - `Fortify::confirmTwoFactorUsing()` - To override the two factor confirmation process, if needed
 - `Fortify::isTwoFactorEnabledUsing()` - To override how to determine if a user has 2FA enabled, if needed
 - `Fortify::isTwoFactorDisabledUsing()` - To override how to determine if a user has 2FA disabled, if needed

Usage Example (`app/Providers/FortifyServiceProvider.php`):

```php
// required
Fortify::confirmTwoFactorView(function () {
    return view('auth.confirm-two-factor');
});

// optional (if using two-factor.confirm)
Fortify::userTwoFactorNotEnabledView(function () {
    return view('auth.user-two-factor-not-enabled');
});

// Optional: Override the confirm two factor handling
Fortify::confirmTwoFactorUsing(function() {
    // code to handle two factor confirmation differently
});
```

### Config:

Fortify's `config/fortify.php`:

 - adds `login_confirms_password` (bool)
 - adds `login_confirms_two_factor` (bool)
 - adds `number_of_recovery_codes` (int)

Laravel's `config/auth.php`:

Add the following to your project's auth config (optional, defaults to 900 without it):

```php
/*
|--------------------------------------------------------------------------
| Two Factor Confirmation Timeout
|--------------------------------------------------------------------------
|
| Here you may define the amount of seconds before a two factor confirmation
| times out and the user is prompted to re-enter their tow factor code via
| the confirmation screen. By default, the timeout lasts for 15 minutes.
|
*/

'two_factor_timeout' => 900,
```

`login_confirms_password` and `login_confirms_two_factor` allows the system to treat the user's login as a
confirmation. This prevents the user from logging in and immediately being asked for their password or two
factor code if they try to access a protected page. That is really annoying. It is disabled by default.

### Middleware

 - adds `confirm.required` middleware
   - requires adding `'confirm.required' => \Laravel\Fortify\Http\Middleware\ConfirmRequired::class` to `app/Kernal.php`
 - adds a `two-factor.confirm` middleware (optional)
   - requires adding `'two-factor.confirm' => \Laravel\Fortify\Http\Middleware\RequireTwoFactor::class` to `app/Kernal.php`

Adding `confirm.required` is necessary for two factor protected routes and is the intended middleware because it will
check if the user has two factor enabled. If so, they will be prompted to enter their two factor code. If not, then
they will be prompted to enter their password. It will use each middleware's respective timeouts.

However, `two-factor.confirm` will only work with two factor, which why the `userTwoFactorNotEnabledView` helper method exists. If you wanted to protect a page and ONLY allow two factor confirmation to it, you could use this. If the user doesn't have two factor enabled, they will be presented with a page that could say, "Please enable two factor
authentication to access this page". I don't know why you would want to do this, but the class already existed, so who
knows..

### Model trait

 - adds helper methods to `TwoFactorAuthenticatable` trait for your model (ie: User)
   - `isTwoFactorEnabled()`
   - `isTwoFactorDisabled()`

These can be overrode in their `FortifyServiceProvider` using the callback methods `Fortify::isTwoFactorEnabledUsing()` and `Fortify::isTwoFactorDisabledUsing()`, allowing the developer to change what constitutes 2fa being enabled or disabled. Plus, it makes the blade template easier to read on the enable/disable page. Example:

```php
@if($user->isTwoFactorEnabled())
    // show QR, codes, and disable button
@else
    // show enable button
@endif
```

### Events

Adds the following helpful events:

 - `PasswordChanged`
 - `ProfileUpdated`
 - `TwoFactorDisabled`
 - `TwoFactorEnabled`

### Docs

The Fortify page on the Laravel docs would need to be updated.

### Still TODO

There is an open issue where a user can be locked out of their account if they don't actually add it to their authenticator app. @taylorotwell said he was going to work on it. Just curious what the status is? My take was to use the session so we don't have invalid data laying in the database. Then if they didn't get it added and their session expires (or they logoff), it is like it never happened.

There is another issue, where if enabling 2fa requires password confirmation, it doesn't enable 2fa the first time. Password confirmation is enabled by default (following the docs). You click enable, confirm your password, and get taken back to the original page. The POST request doesn't make it through the confirmation. Once you confirm, you can click enable (again) and it will go through the second time. I am not really sure how to solve this other than to require password confirmation on the page itself instead of on the POST method.

### Demo

I have created a demo starter project that showcases these modifications. By default, it is running Fortify with the default settings. Following the Readme, you can enable and use the modifications.

Github Repo: [Laravel Fugly Starter](https://github.com/WadeShuler/laravel-fugly-starter)


driesvints:
Putted this in draft until you've finished up the PR 👍

WadeShuler:
I'll take a look at the error check. Looks like I see why the password rules were duplicated in there lol.

WadeShuler:
@driesvints I think I got it down to 1 error and 1 style ci complaint...

I am having a hard time getting the test to generate a recovery code, store that code in a variable, and pass it back through the check. `Laravel\Fortify\Tests\NewPasswordControllerTest::test_the_new_password_view_is_returned`

In `src/Http/Controllers/NewPasswordController`:

```php
public function create(Request $request): ResetPasswordViewResponse
{
    $user = $this->broker()->getUser(
        $request->only(Fortify::email())
    );

    if (! $user || ! $exists = $this->broker()->tokenExists($user, $request->route('token'))) {
        abort(401, 'Invalid Token!');
    }

    return app(ResetPasswordViewResponse::class);
}
```

As you can see, it checks if the token is valid, else throws a 401 as per specs. The test tries to use a fake token.. So I either need to get it to accept a fake token.. or to generate a real token and pass it in the url in the test. I haven't done much Laravel testing, so I don't know how to remedy this without spending hours lol. Any guidance would be appreciated.

I don't know what the StyleCI is complaining about. It highlights `use Laravel\Fortify\Http\Middleware\RequireTwoFactor;` in `src/Http/Middleware/ConfirmRequired.php`.. but `RequireTwoFactor` is used. So does it not want me to include it via `use` because it's in the same namespace path?

I already fixed the other StyleCI complaint. Just going to push them both at the same time.

WadeShuler:
@driesvints - The demo project is done. All you have to do is clone it and edit the env as you would any new Laravel project (db, mail, etc). It is basically a default Fortify setup. Some of the fixes are built in, like redirecting back to the code page on an invalid code, valid verification token checking, etc. That's unavoidable as they are built-in fixes.

Following the README will walk you through enabling the new features. Things like `login_confirms_password` and `login_confirms_two_factor`, and changing the number of recovery codes. Register a new user, login, checkout the links in the top nav. Enable 2FA and check out the "2FA Protected" and "protected" pages. You will notice the "2FA Protected" page will tell you that you don't have 2fa enabled. The "Protected" page will require your password. If you enable 2FA in settings.. log out and back in.. Then try the "Protected" page again, it will this time ask for your 2fa code because the user has enabled 2fa. If you follow the readme to enable `login_confirms_password` and `login_confirms_two_factor`, then logout and back in, you will notice it doesn't ask for you to confirm when you try to access those pages. By default, password confirmation is set to 1 hour and 2fa confirmation is set to 15 minutes. If you wait past that time, it will require you to confirm.

Demo Project: [Laravel Fugly Starter](https://github.com/WadeShuler/laravel-fugly-starter)

driesvints:
@WadeShuler taylor will review this one 👍

taylorotwell:
I'm going to keep it real - I don't have time to review and vet all this. I think I stated on another thread that if I want 2FA confirmation I will be implementing it myself on Fortify. I just don't have the bandwidth to review PRs this large on a security related project. Take care!

WadeShuler:
@taylorotwell Thank you for your prompt reply, I understand. In the other thread you mentioned fixing the "2FA can lock user out" issue, which is why I didn't touch that.

I think if the 2fa used pipelines like the normal login, it would really help and allow us the flexibility to do whatever we need.

I am sure you can code it 10x better than I could, so I look forward to seeing it.

Issue #343, opened 2022-01-14
2FA invalid code redirects back to login


talelmishali:
- Fortify Version: 1.9.0
- Laravel Version: 8.79.0
- PHP Version: 8.1.1
- Database Driver & Version: MySql 8.0.27

### Description:
With the current implementation when a user provides an invalid 2FA code, the user is been redirected to the login route. Instead, a more suitable approach would be to redirect the user back to the 2FA route for re-entering the code (same as the behavior on the forge website and many others).

Before opening a PR I wanted to know if it is something the laravel team would consider addressing at the moment, I know it was mentioned on a previous PR (#325 ) and this is why I am asking (@taylorotwell mentioned he might not have the time to vet the code at moment).

### Steps To Reproduce:
1. Log in to an account with the 2FA feature activated 
2. Enter invalid 2FA code on verification

alexandercrice:
@talelmishali In case there is no 

```php
$this->app->extend(
    \Laravel\Fortify\Contracts\FailedTwoFactorLoginResponse::class,
    function ($response, $app) {
        return new class($response) implements \Laravel\Fortify\Contracts\FailedTwoFactorLoginResponse {
            private $baseResponse;

            public function __construct(\Laravel\Fortify\Contracts\FailedTwoFactorLoginResponse $response)
            {
                $this->baseResponse = $response;
            }

            public function toResponse($request)
            {
                $response = $this->baseResponse->toResponse($request);
                if ($response instanceof \Illuminate\Http\RedirectResponse) {
                    $response->setTargetUrl(
                        // ... the URL you want to redirect to
                    );
                }
                return $response;
            }
        };
    }
);

alexandercrice:
@talelmishali 

I ran into a similar problem. In case there is no fix coming for this, here is a workaround I used when I needed to customize redirect behavior for responses that don't look at the fortify.php configuration file. I haven't tried the 2FA feature yet, but it looks like the same approach would work. Use the `extend` method to bind a new response class that wraps/decorates (whatever you want to call it) the implementation response (`Laravel\Fortify\Http\Responses\FailedTwoFactorLoginResponse` ) from the Fortify package. All this does is it takes the response object returned by the FailedTwoFactorLoginResponse's `toResponse` method, and changes the URL before passing it along.

```php
$this->app->extend(
    \Laravel\Fortify\Contracts\FailedTwoFactorLoginResponse::class,
    function ($response, $app) {
[...]

PR #344, opened 2022-01-17
[2FA] Entering an invalid 2FA code redirects back to re-enter the code

talelmishali:
This PR solves #343 
#### Changes
- Entering an invalid 2FA code results in a redirect back to allow the user to re-enter the code
- More meaningful error message when a user enters an invalid recovery code vs 2FA code

The benefit to the end-users would be a better UX in my opinion as described previously in #343

### DIFF
--- a/src/Http/Controllers/TwoFactorAuthenticatedSessionController.php
+++ b/src/Http/Controllers/TwoFactorAuthenticatedSessionController.php
@@ -63,7 +63,7 @@ class TwoFactorAuthenticatedSessionController extends Controller
         } elseif (! $request->hasValidCode()) {
             $request->session()->put([
                 'login.id' => $user->getKey(),
-                'login.remember' => $request->filled('remember'),
+                'login.remember' => $request->remember(),
             ]);
 
             return app(FailedTwoFactorLoginResponse::class);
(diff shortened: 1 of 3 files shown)


src/Http/Controllers/TwoFactorAuthenticatedSessionController.php
```diff
@@ -61,6 +61,11 @@ public function store(TwoFactorLoginRequest $request)
 
             event(new RecoveryCodeReplaced($user, $code));
         } elseif (! $request->hasValidCode()) {
+            $request->session()->put([
+                'login.id' => $user->getKey(),
+                'login.remember' => $request->filled('remember'),
```
taylorotwell, line 66: Why not `$request->remember()`?

talelmishali: @taylorotwell you're right, at this point ```$request->filled('remember')``` not going to help. It needs to be ```$request->remember()```.

taylorotwell:
I dunno - I don't think I'm going to change this right now.

talelmishali:
Ok, thanks.

PR #346, opened 2022-01-19
Customisable login validation rules

andreaselia:
This takes the currently password trait implementation and does the same again but for the login validation rules. There's not a whole lot to it as a PR, but it does increase the flexibility of people wanting to change the validation rules since the LoginRequest cannot be overridden.


taylorotwell:
This isn't done correctly I don't think. What if the user's `App` namespace isn't `App`?

andreaselia:
Cheers @taylorotwell, I'll look for another way of achieving this 👍

andreaselia:
@taylorotwell after looking through the rest of Fortify, it seems as though the `App` namespace is referenced through the package, with one example being here:

https://github.com/laravel/fortify/blob/1.x/stubs/PasswordValidationRules.php#L3

andreaselia:
Do you have an update on the above @taylorotwell?

phpmac:
i need this one

Issue #351, opened 2022-02-03
Two Factor Authentication returns error on subsequent code entries if first one was invalid


devstack-be:
- Fortify Version: 1.7
- Laravel Version: 8.40
- PHP Version: 8
- Database Driver & Version: MYSQL

### Description:
I am building an API by using Laravel 8 with Fortify and with vueJS on frontend. I enabled and integrated two-factor authentication and it works well, but just stumbled upon an issue which I do not know if it's by design or a bug.

So, the problem is when user is logging in and he enters the 2fa code wrongly for the first time. In this case, on all subsequent entries, even if the code is correct and valid, it will reject it. After some digging around, I noticed that the login.id stored in session in Laravel\Fortify\Actions\RedirectIfTwoFactorAuthenticatable::twoFactorChallengeResponse is cleared out if the code entered first time is invalid. On all subsequent requests, the login.id in Laravel\Fortify\Http\Requests\TwoFactorLoginRequest::challengedUser() is returning null.

So, my question is: is this intentional and by design? Or did I discover a bug?

My reasoning is that this is a bug because user already provided username and password, now he only needs to enter valid 2fa code to login completely, so not sure why forcing user to enter username and password again if they typed 2fa code wrongly for the first time.

The issue happens in Laravel\Fortify\Http\Requests\TwoFactorLoginRequest::challengedUser() where the code executes $model::find($this->session()->pull('login.id')). pull() is designed to get an item from session storage and then forget it immediately after. Once I changed that into get(), it started working as it should.

driesvints:
Hey @devstack-be. So I believe this to be intentionally but agree it's a bit odd as that's not the behavior of most systems. 

I've attempted a PR here: https://github.com/laravel/fortify/pull/353. I'm going to close this one as it's more a request to change the behavior here rather than an actual bug. Let's see how the PR goes. 

Thanks

devstack-be:
@driesvints Thanks for your quick answer.

PR #353, opened 2022-02-04
[1.x] Prevent new login after failed 2FA

driesvints:
At the moment, when a 2FA challenge is failed, the user is required to again provide its username and password. This, however, isn't standard behavior across most apps. Usually, you're not required to login again and are asked to again attempt a new 2FA code.

I've attempted a fix that keeps the `login.id` in session after a failed 2FA and the failed 2FA response now redirects to the 2FA form. The `login.id` key is only removed after a successful 2FA code. Together with the 2FA rate limiting I feel this should provide enough safety. I don't really see how asking the username/password again for every failed 2FA request is more safer.

If this is considered a breaking change then I can send this to master. 

See https://github.com/laravel/fortify/issues/351

### DIFF
--- a/src/Http/Requests/TwoFactorLoginRequest.php
+++ b/src/Http/Requests/TwoFactorLoginRequest.php
@@ -54,9 +54,13 @@ class TwoFactorLoginRequest extends FormRequest
      */
     public function hasValidCode()
     {
-        return $this->code && app(TwoFactorAuthenticationProvider::class)->verify(
+        return $this->code && tap(app(TwoFactorAuthenticationProvider::class)->verify(
             decrypt($this->challengedUser()->two_factor_secret), $this->code
-        );
+        ), function ($result) {
+            if ($result) {
+                $this->session()->forget('login.id');
+            }
+        });
     }
 
     /**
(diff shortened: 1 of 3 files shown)


sebastiaanluca:
We don't use all of Fortify's features (just login and 2FA, but with custom routing), so this was a gotcha our tests luckily caught. Seems like a big change for a patch release 😬

driesvints:
@sebastiaanluca how can this affect you if you were overwriting the responses already?

sebastiaanluca:
@driesvints Weren't overwriting anything (except providing our own views), only disabling Fortify's routes and adding our own for just the login and 2FA. Those were using Fortify's controllers and logic. Yet we were depending on the 2FA form to redirect back to the login page after an unsuccessful attempt.

It's not a big thing, just a caveat that can cause issues in some apps and not at all expected in a patch release. Makes sense to just show the same 2FA form again instead of redirect back to the login page 👍

driesvints:
@sebastiaanluca I have a day off today but I'll have another look at this tomorrow. In any case, thanks for making me aware 👍

driesvints:
> adding our own for just the login and 2FA.

Did the routes you add have the same names as the ones from fortify?

sebastiaanluca:
Nope, that was one of the causes of the issue, alongside our 2FA form not showing form errors (because they were shown on the login page after redirect). But I fixed it in our app when updating the package, so all is good now.

driesvints:
Ah okay. Because if you do overwrite the routes then it's important that you preserve the route names. That's always necessary for the package to function properly.

PR #357, opened 2022-02-17
Confirmable 2FA

taylorotwell:
This PR allows 2FA to be configured to require confirmation before it is fully enabled via a new feature option. I don't *think* this PR is even breaking? But, sending to `master` for now.


driesvints:
I feel like we should be able to do this in Fortify v1. By default `Features::optionEnabled(Features::twoFactorAuthentication(), 'confirm')` will return `false` so it should not break any existing apps. 

However, in Fortify 2.x I would set it to be enabled by default.

WadeShuler:
A very necessary PR to make Fortify usable with real-world customers 👍

PR #358, opened 2022-02-18
[1.x] Confirmable 2FA

driesvints:
Replaces https://github.com/laravel/fortify/pull/357

### DIFF
--- a/database/migrations/2014_10_12_200000_add_two_factor_columns_to_users_table.php
+++ b/database/migrations/2014_10_12_200000_add_two_factor_columns_to_users_table.php
@@ -22,7 +22,9 @@ return new class extends Migration
                     ->after('two_factor_secret')
                     ->nullable();
 
-            $table->boolean('two_factor_confirmed')->default(false);
+            $table->boolean('two_factor_confirmed')
+                    ->after('two_factor_recovery_codes')
+                    ->default(false);
         });
     }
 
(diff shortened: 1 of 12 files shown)


src/Actions/RedirectIfTwoFactorAuthenticatable.php
```diff
@@ -50,6 +50,16 @@ public function handle($request, $next)
     {
         $user = $this->validateCredentials($request);
 
+        if (Fortify::confirmsTwoFactorAuthentication()) {
+            if (optional($user)->two_factor_secret &&
+                optional($user)->two_factor_confirmed &&
+                in_array(TwoFactorAuthenticatable::class, class_uses_recursive($user))) {
+                return $this->twoFactorChallengeResponse($request, $user);
+            } else {
+                return $next($request);
```
fkeloks, line 59: This `else` is not necessary thanks to the `return` of the `if` above :)

database/migrations/2014_10_12_200000_add_two_factor_columns_to_users_table.php
```diff
@@ -21,6 +22,12 @@ public function up()
             $table->text('two_factor_recovery_codes')
                     ->after('two_factor_secret')
                     ->nullable();
+
+            if (Fortify::confirmsTwoFactorAuthentication()) {
+                $table->boolean('two_factor_confirmed')
```
Jubeki, line 27: Just out of convention like:  `email_verified_at` wouldn't it be better to use a timestamp instead of a boolean? This would also allow it to show when the 2fa was activated.

```suggestion
                $table->timestamp('two_factor_confirmed_at')
```

Jubeki: I propose: #359

PR #368, opened 2022-03-04
[1.x] Ensures route password.confirm is defined when not using views

### LINKED ISSUE #367: Missing route name for password.confirm without enabled views
Fortify Version: any
Laravel Version: any
PHP Version: any

### Description:
If you implement fortify without views, you have to hardcode the `password.confirm` route because no name is defined. 

https://github.com/laravel/fortify/blob/b2de125809a742acf08bc6bcbe8f90f8ae92df29/routes/routes.php#L120-L121

If you have views enabled, you are saved because a get route is defined with the same route.

https://github.com/laravel/fortify/blob/b2de125809a742acf08bc6bcbe8f90f8ae92df29/routes/routes.php#L110-L114

I am not sure of a good way to fix this, if you give this route a name, it will break existing implementations; a solution might be to do something like the following until a major version can be released to clean it up.

```php
if ($enableViews) {
    Route::get('/user/confirm-password', [ConfirmablePasswordController::class, 'show'])
        ->middleware([config('fortify.auth_middleware', 'auth').':'.config('fortify.guard')])
        ->name('password.confirm');

    Route::post('/user/confirm-password', [ConfirmablePasswordController::class, 'store'])
        ->middleware([config('fortify.auth_middleware', 'auth').':'.config('fortify.guard')]);
} else {
    Route::post('/user/confirm-password', [ConfirmablePasswordController::class, 'store'])
        ->middleware([config('fortify.auth_middleware', 'auth').':'.config('fortify.guard')])
        ->name('password.confirm');
}
```

While removing the code here

https://github.com/laravel/fortify/blob/b2de125809a742acf08bc6bcbe8f90f8ae92df29/routes/routes.php#L120-L121

Frozire:
This fixes #367 

If you implement fortify without views, you have to hardcode the `password.confirm` route because no name is defined on the `post` route.

https://github.com/laravel/fortify/blob/b2de125809a742acf08bc6bcbe8f90f8ae92df29/routes/routes.php#L120-L121

If you have views enabled (`$enableViews = true`), you are saved because a `get` route is defined with the same route.

https://github.com/laravel/fortify/blob/b2de125809a742acf08bc6bcbe8f90f8ae92df29/routes/routes.php#L110-L114

Currently, this forces people using Fortify with a SPA/custom views (`$enableViews = false`), to hardcode the `password.confirm` route.

While this might not be the prettiest fix, it is a fix that will not cause breaking changes for those already using the package.

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -109,20 +109,16 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
     // Password Confirmation...
     if ($enableViews) {
         Route::get('/user/confirm-password', [ConfirmablePasswordController::class, 'show'])
-            ->middleware([config('fortify.auth_middleware', 'auth').':'.config('fortify.guard')])
-            ->name('password.confirm');
-
-        Route::post('/user/confirm-password', [ConfirmablePasswordController::class, 'store'])
             ->middleware([config('fortify.auth_middleware', 'auth').':'.config('fortify.guard')]);
-    } else {
-        Route::post('/user/confirm-password', [ConfirmablePasswordController::class, 'store'])
-            ->middleware([config('fortify.auth_middleware', 'auth').':'.config('fortify.guard')])
-            ->name('password.confirm');
     }
 
     Route::get('/user/confirmed-password-status', [ConfirmedPasswordStatusController::class, 'show'])
         ->middleware([config('fortify.auth_middleware', 'auth').':'.config('fortify.guard')])
         ->name('password.confirmation');
+
+    Route::post('/user/confirm-password', [ConfirmablePasswordController::class, 'store'])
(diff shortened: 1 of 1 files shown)


driesvints:
Another easier solution is to just name the post route and remove it from the get route.

Frozire:
> Another easier solution is to just name the post route and remove it from the get route.

Wouldn't that cause the get route to not be callable?

driesvints (approved):

bessone:
with this modification I get an error with the Laravel Breadcrumbs project (https://github.com/diglactic/laravel-breadcrumbs), all the routes in GET must have a name to be used with the breadcrumbs, having moved the name from GET to POST there is no more way to make it work.

PR to correct use without views has created the problem for those who use views (with breadcrumbs)

PR #360, opened 2022-02-20
Fix incorrect key for error bag.

vaibhavpandeyvpz:
The key for the error bag if 2FA authentication attempt was unsuccessful is incorrectly passed here.

PR #369, opened 2022-03-17
[1.x] Fix double error message for failed 2FA response

driesvints:
This PR solves an issue where a failed message was shown twice. This is technically a breaking change since we're removing the email key (the reason why we didn't in the related PR below). But the correct key is `code`.

See https://github.com/laravel/fortify/pull/360


sebastiaanluca:
Thanks for the fix 🙌 , though please mention these breaking changes in the changelog 😅  We're always missing them for Fortify because they're included in minor releases, resulting in views not working or error messages not being displayed. In this case we were only displaying the error for `email`, otherwise we had duplicate error messages like in the Jetstream example.

driesvints:
@sebastiaanluca yes sorry about that. We really try to minimize these... unfortunately this was a necessary one.

PR #371, opened 2022-03-18
2FA setup key

ps-sean:
* Added new route & controller to allow users to use setup key when enabling 2FA

This is necessary for jetstream to be able to support showing the setup key when using inertia


magdicom:
wouldn't be better if the setup key is placed in the same page as the qrcode, maybe below it?

ps-sean:
> wouldn't be better if the setup key is placed in the same page as the qrcode, maybe below it?

That’s how I’ve done it in Jetstream. This PR is purely for the inertia build as it requires an API endpoint in fortify to pull the key through.

PR #373, opened 2022-03-20
fix optional feature: Two-Factor confirmation

magdicom:
after installation completed, and once you run the `migrate` command, enabling the Two-Factor confirmation feature will break the Two-Factor "Disable" button at least.

Column not found: 1054 Unknown column 'two_factor_confirmed_at' in 'field list'

### DIFF
--- a/database/migrations/2014_10_12_200000_add_two_factor_columns_to_users_table.php
+++ b/database/migrations/2014_10_12_200000_add_two_factor_columns_to_users_table.php
@@ -23,11 +23,9 @@ return new class extends Migration
                     ->after('two_factor_secret')
                     ->nullable();
 
-            if (Fortify::confirmsTwoFactorAuthentication()) {
-                $table->timestamp('two_factor_confirmed_at')
-                        ->after('two_factor_recovery_codes')
-                        ->nullable();
-            }
+            $table->timestamp('two_factor_confirmed_at')
+                    ->after('two_factor_recovery_codes')
+                    ->nullable();
         });
     }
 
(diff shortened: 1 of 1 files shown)


driesvints:
The current code is correct. You need to write your own migration for this if you're upgrading.

ChrisRiddell:
> The current code is correct. You need to write your own migration for this if you're upgrading.

For new installs, This should be documented in the docs that you need to enable the feature before running the migration.

driesvints:
@ChrisRiddell yes I agree. Can you send in a PR?

magdicom:
> The current code is correct. You need to write your own migration for this if you're upgrading.

it was a new installation, all I did was to make some changes to the /config/fortify.php, just to discover this issue.

instead, it should be mentioned in the installation guide, that you need to make your configuration before running the `migrate` command, or provide clear info about this particular feature and a migration code to run if the developer want it enabled.

driesvints:
@magdicom yes that's what @ChrisRiddell already noted. Can anyone send in a pr to the docs?

magdicom:
> @magdicom yes that's what @ChrisRiddell already noted. Can anyone send in a pr to the docs?

my point is, if that is an optional feature, it would be better if the required database column is there in the first place.

that's why I sent this pr.

driesvints:
We can't add the database column as required because it would break existing apps.

ChrisRiddell:
> @ChrisRiddell yes I agree. Can you send in a PR?

I noticed the confirm functionality is not documented yet, Happy to have a go at doing a PR for it tomorrow.

driesvints:
@ChrisRiddell Think Taylor will do that indeed tomorrow. So maybe wait until he has done that.

taylorotwell:
Not sure this is necessary. Soon `confirm` will be the default.

For existing applications that want to enable this feature then yes some custom migration will be needed.

PR #387, opened 2022-05-11
Add two factor authentication successful event

Silver343:
This PR Adds a Two factor Authentication Success event,

This is useful if apps want to keep a record of the last time a two factor prompt was successful.


taylorotwell:
I think you can use the existing `Login` event provided by Laravel and then just check if the user instance attached to that event has two factor authentication enabled and confirmed.

Issue #388, opened 2022-05-12
Return a Responsable interface instance instead of pre-defined JsonResponse or RedirectResponse on update profile action


a-bashtannik:
`ProfileInformationController::update` [can return 2 kinds of response instances](https://github.com/laravel/fortify/blob/1.x/src/Http/Controllers/ProfileInformationController.php#L24) only: `JsonResponse` or `RedirectResponse`. 

In the same time, `RegisteredUserController::store` [allows customization by returning](https://github.com/laravel/fortify/blob/1.x/src/Http/Controllers/RegisteredUserController.php#L58) resolved `RegisterResponse` which behavior can be redefined somewhere in the app's service providers by developers via IoC. 

Same approach applies in other controllers like `PasswordResetLinkController` or `NewPasswordController`. 

I suggest changing the behavior of `ProfileInformationController::update`, extracting the method implementation, and returning resolved `UpdateProfileResponse` instance, like in other examples, allowing developers customizing the response if need. 

I'm ready to create a pull request.

driesvints:
Sure, feel free to attempt a PR.

PR #389, opened 2022-05-13
[1.x] Return a Responsable instance on profile update (#388)

a-bashtannik:
Allows developers to customize response on update profile action via IoC like it was done in other user experience flows (e.g. Register). 

This PR doesn't bring backward incompatibility because `\Laravel\Fortify\Http\Responses\UpdateProfileResponse` class implements the previous testable behavior.

### DIFF
--- a/src/Contracts/UpdateProfileResponse.php
+++ b/src/Contracts/UpdateProfileResponse.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Laravel\Fortify\Contracts;
+
+use Illuminate\Contracts\Support\Responsable;
+
+interface UpdateProfileResponse extends Responsable
+{
+    //
+}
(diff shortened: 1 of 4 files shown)


a-bashtannik:
Hey @taylorotwell 

This update brings a really good approach by introducing some inversion of control, allowing the developer to return a complete user object in the update response, like in my case - we need to return the object on every update operation. Without this PR it's impossible because of the hard-coded JSON empty string response.

Other [13 cases use such an approach](https://github.com/laravel/fortify/blob/1.x/src/FortifyServiceProvider.php#L71), so I'm wondering why do you prefer to keep update profile action flow another.

a-bashtannik:
Hey @driesvints , could you share our opinion, please?

driesvints:
@a-bashtannik I guess Taylor doesn't wants to merge this in right now, sorry.

PR #398, opened 2022-07-26
Add at lest count number to password rule with backward compatibility

sajjadRabiee:
Add at lest count to password rule with backward compatibility for version one

- atLestNumericCount => default is one
- atLestUpperCaseCount => default is one
- atLestSpecialCharacterCount => default is one


sajjadRabiee:
thanks @taylorotwell for close my pull request I think you have a reason for this but this is my first try to contribute in an open source project if you have an advice for me I'm glade to hear that

PR #399, opened 2022-08-08
Fix second usage of 2FA code

xwillq:
Method `verifyKeyNewer` returns `true` when parameter `$oldTimestamp` is empty i.e. when there is no such code in cache. This allowed second usage of 2FA code:

#### First login with 2FA code:
`verifyKeyNewer` gets called with `null` as old timestamp, so it returns `true`. `true` is saved to cache as old timestamp.
#### Second login with same 2FA code:
`verifyKeyNewer` gets called with `true` as old timestamp, which later down the line gets converted to `1`. `verifyKeyNewer` uses current time and window value for creating starting timestamp and discards old timestamp, because it is smaller than created timestamp. So, if 2FA code is within the window, it gets accepted.

This pull request fixes this bug by creating new timestamp if `verifyKeyNewer` returned `true` and writing this new timestamp to cache.

### DIFF
--- a/src/TwoFactorAuthenticationProvider.php
+++ b/src/TwoFactorAuthenticationProvider.php
@@ -79,6 +79,7 @@ class TwoFactorAuthenticationProvider implements TwoFactorAuthenticationProvider
             if ($timestamp === true) {
                 $timestamp = $this->engine->getTimestamp();
             }
+
             optional($this->cache)->put($key, $timestamp, ($this->engine->getWindow() ?: 1) * 60);
 
             return true;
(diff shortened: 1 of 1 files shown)


driesvints:
Doesn't this defeats the purpose of the "one" in OTP?

xwillq:
Maybe description isn't clear enough (I'm not a native speaker, sorry), but **_current_** implementation allows using code twice. This pull request fixes the bug.

xwillq:
To reproduce the bug you just need to use code second time while it is active, i.e `30 * $window` seconds from code generation.

driesvints:
Got it 👍

PR #400, opened 2022-08-10
Registration form request

joelbutcher:
This PR moves from an array based input for user creation to an interface driven, from request approach. I've also defined a new, publishable trait for username validation rules (similar to password validation, this is because the username field may not be an `email` field as defined in the Fortify config file).

I've made this change because the current approach requires developers to step away from a common Laravel convention and requires knowledge (and some documentation crawling) to find out that registration validation is held within a class who's primary "named responsibility" is to create something (which is somewhat confusing). This approach takes inspiration from Laravel.io's usage of jobs to create objects from form requests. The idea being that an object / interface is easier to work with and provides developers more visibility into what the code is doing / intended for. 

Secondly, it allows developers to separate the concern of validation from the action (again, which is named for "creation"). This separation of concerns allows for better visibility into what parts of the application are responsible for what.

Users can override the concrete implementation by updating the `Fortify::registerUsersVia(...)` call in the published service provider.

---

**Additional notes:**
1. Since the `name` field may not be required by some developers, I've added the `null` return type to docblock type hints in the new contract and the current implementation. Developers can simply `return null` in this method and remove the validation rule from the `rules()` method if this is not required for their use-case.
2. I've added a new `UsernameValidationRules` trait for developers to define the specific validation rules required for a username  / email. Ideally, this should also be used for validating the `Fortify::username()` field in `LoginRequest::rules()`. 
3. If this is accepted, we will need to consider this change in the next version of Laravel Jetstream

### DIFF
--- a/src/Contracts/CreatesNewUsers.php
+++ b/src/Contracts/CreatesNewUsers.php
@@ -2,14 +2,12 @@
 
 namespace Laravel\Fortify\Contracts;
 
-use App\Http\Requests\UserRegistrationRequest;
-
 interface CreatesNewUsers
 {
     /**
      * Validate and create a newly registered user.
      *
-     * @param  \App\Http\Requests\UserRegistrationRequest  $request
+     * @param  \Laravel\Fortify\Contracts\UserRegistrationRequest  $request
      * @return mixed
      */
     public function create(UserRegistrationRequest $request);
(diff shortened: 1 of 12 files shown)


taylorotwell:
I don't think I really want to change things here at this time.

joelbutcher:
A bit of a shame, I'll probably maintain a fork for this, but I do appreciate the time taken to review, regardless of the outcome. Thanks, Taylor!

phpmac:
i need this one,

PR #401, opened 2022-08-12
[1.x] Return recovery errors under the `recovery_code` key

jessarcher:
_This might be considered a breaking change..._

Currently when the user sends a `recovery_code`, any basic validation errors will be returned under the `recovery_code` key. But if the provided recovery code is incorrect, then the error is returned under the `code` key instead.

It's not a big deal if all errors are displayed at the top of the page. But if the errors are displayed alongside the specific fields (like [this Jetstream PR](https://github.com/laravel/jetstream/pull/1115#discussion_r941912210)), then the developer really needs to look in two places for recovery code errors.

This PR returns all recovery code errors under the `recovery_code` key.

The only time this may be a problem is if the user has worked around this quirk by *only* checking the `code` key for `recovery_code` errors. If this is considered a breaking change, then I can target `master` instead.

### DIFF
--- a/src/Http/Responses/FailedTwoFactorLoginResponse.php
+++ b/src/Http/Responses/FailedTwoFactorLoginResponse.php
@@ -15,14 +15,16 @@ class FailedTwoFactorLoginResponse implements FailedTwoFactorLoginResponseContra
      */
     public function toResponse($request)
     {
-        $message = __('The provided two factor authentication code was invalid.');
+        [$key, $message] = $request->has('recovery_code')
+            ? ['recovery_code', __('The provided two factor recovery code was invalid.')]
+            : ['code', __('The provided two factor authentication code was invalid.')];
 
         if ($request->wantsJson()) {
             throw ValidationException::withMessages([
-                'code' => [$message],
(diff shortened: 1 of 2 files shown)


driesvints:
Won't this break Laravel apps that update Fortify but not Jetstream? Also, we indeed had complaints in the past from people implementing Fortify themselves when we changed some error code keys. I feel it might be better to target master for this.

jessarcher:
I don't believe this will break existing Jetstream installs because they display _all_ validation errors at the top of the page. The key isn't used afaik.

To me, this feels more like a bug fix. But, as you say, people may be depending on the `code` key for their `recovery_code` requests.

Note that this won't change the `code` key for standard 2FA logins. It's just recovery logins that send a bad code that may not see an error, depending on their customisations.

driesvints:
@sebastiaanluca sorry about that... I was afraid you'd get caught again 😞 

>  so could breaking changes like this be avoided if we just use Jetstream/Breeze + Fortify

Not really since Jetstream is a scaffold and existing views in apps don't get updated.

PR #415, opened 2022-10-26
Fix error message when entering invalid 2fa code

emargareten:
Currently when user enters an invalid code and an empty recovery code (this is how jetstream sends the code) the error message key will always be `recovery_code` (and won't display in jetstream).


jessarcher (approved):
Thanks!

Issue #440, opened 2023-02-15
Password Rule implements deprecated Illuminate\Contracts\Validation\Rule


falko100:
- Fortify Version: v1.16.0
- Laravel Version: v10.0.0
- PHP Version: 8.1
- Database Driver & Version: n.a.

### Description:
`\Laravel\Fortify\Rules\Password`
implements deprecated `Illuminate\Contracts\Validation\Rule`
instead of `Illuminate\Contracts\Validation\ValidationRule`

#### Edit:
I won't mind giving this a go for contributing. Do you have any example of how you do a test for the `Illuminate\Contracts\Validation\ValidationRule` ?

falko100:
Would this be the way to go?
```php
$this->assertThrows(function () {
    validator(
        ['password' => 'secret'],
        ['password' => [new Password]],
    )->validate();
}, ValidationException::class, 'The password must be at least 8 characters.');
```

driesvints:
I'm not sure myself. Maybe try discord to see if there's someone who can help you with the PR?

PR #441, opened 2023-02-15
Implement ValidationRule instead of ValidationRule at Rules/Password

falko100:
See https://github.com/laravel/fortify/issues/440

### DIFF
--- a/src/Rules/Password.php
+++ b/src/Rules/Password.php
@@ -2,10 +2,10 @@
 
 namespace Laravel\Fortify\Rules;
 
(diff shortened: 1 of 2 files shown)


falko100:
Since the deprecation is added in Laravel 10, I probably just gotta ignore the phpstan depraction error for a while 😅

johanmolen:
@falko100 I can test this solution if you want?

falko100:
> @falko100 I can test this solution if you want?

Sure thing!

PR #446, opened 2023-03-04
[1.x] Add registerThrough Callback to add custom Middlewares to Registration Route

stefanzweifel:
In a project I want to use the [spatie/laravel-honeypot](https://github.com/spatie/laravel-honeypot) package to protect the app against spam sign ups.
However, there currently doesn't seem to be a way to configure the middlewares of the `POST /register`-route.

This PR adds a `registerThrough()`-method and `$registerThroughCallback`-property to `Fortify`, that works similar to the [`authenticateThrough()`-method](https://laravel.com/docs/10.x/fortify#customizing-the-authentication-pipeline).

User will be able to add middlewares for the registration route by adding the following code to their `FortifyServiceProvider`.

```php
Fortify::registerThrough(function (Request $request) {
    return [
        MyCustomMiddleware::class,
        \Spatie\Honeypot\ProtectAgainstSpam::class
    ];
});
```

> **Warning**
> I had to change the return type of the `store`-method from `Laravel\Fortify\Contracts\RegisterResponse` to `mixed` as my test threw a type error. Not sure why though; my knowledge of the Pipeline class is quite limited.

More than happy to make a PR to the docs repo too, if this get's accepted.

---

After sleeping over this PR, I think it might be overkill to add this to Fortify.
To modify the middlewares for the `/register`-route, all I needed to do was copy & paste the route definition to my user-land routes file.

```php
Route::post('/register', [RegisteredUserController::class, 'store'])
    ->middleware([
        'guest:'.config('fortify.guard'), 
        \Spatie\Honeypot\ProtectAgainstSpam::class
    ]);

// other routes of my app
```

Proposing this pull request anyway, as this might be something you want to see in Fortify anway.

### DIFF
--- a/src/Fortify.php
+++ b/src/Fortify.php
@@ -32,6 +32,13 @@ class Fortify
     public static $authenticateUsingCallback;
 
     /**
+     * The callback that is responsible for building the registration pipeline array, if applicable.
+     *
+     * @var callable|null
+     */
+    public static $registerThroughCallback;
+
+    /**
      * The callback that is responsible for confirming user passwords.
      *
      * @var callable|null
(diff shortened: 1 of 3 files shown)


taylorotwell:
Closing since you found a simple workaround.

PR #451, opened 2023-03-23
Add maxLength for Password validation rule

antja0:
FYI: Our company just received a security vulnerability report about us not having max password length:

> Hashing large data can cause significant resource consumption on behalf of the server expose the app for a DOS attack.

This pull request adds a default max password length for the Fortify Password validation rule.

Note: This change does change the translation keys, so not 100% backwards compatible.


driesvints:
It seems that this breaks the test suite.

PR #452, opened 2023-03-23
Add default max length for passwords

antja0:
Another approach to #451

~~Maybe it won't break the tests this time?~~ It didn't 🥳 

Won't break the translations like the previous PR

FYI: Our company just received a security vulnerability report about us not having max password length:

> Hashing large data can cause significant resource consumption on behalf of the server expose the app for a DOS attack.

### DIFF
--- a/stubs/PasswordValidationRules.php
+++ b/stubs/PasswordValidationRules.php
@@ -13,6 +13,6 @@ trait PasswordValidationRules
      */
     protected function passwordRules(): array
     {
-        return ['required', 'string', new Password, 'confirmed'];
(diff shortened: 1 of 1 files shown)


taylorotwell:
This is a stub file so you can just change it in your own application to whatever you want.

antja0:
Update: I later found this vulnerability is not important.

One reason is that using the PASSWORD_BCRYPT as the algorithm will result in the password parameter being truncated to a maximum length of 72 bytes. https://www.php.net/manual/en/function.password-hash.php

I also was informed by beginning of this video: https://youtu.be/lr1KuL8OmJY?si=0vng-CxjHTSypZaq

PR #457, opened 2023-04-06
[1.x] Add `twoFactorSetupKey` method

jrd-lewis:
This adds a method to access the user's two-factor setup key. This could make a Blade view for setting up 2FA easier to read compared to decrypting the ```two_factor_secret``` field directly in the view.

(Changed to personal repo instead of organization repo, per https://github.com/laravel/fortify/pull/456#issuecomment-1499517339)

**Breaking Changes**
I wasn't sure if it would be a breaking change to edit the ```twoFactorQrCodeUrl``` method to reference this new method; so, I undid that part of the PR on the chance that it would be.

**Tests**
I didn't include any tests; but I could add one if wanted.

### DIFF
--- a/src/TwoFactorAuthenticatable.php
+++ b/src/TwoFactorAuthenticatable.php
@@ -81,7 +81,7 @@ trait TwoFactorAuthenticatable
         return app(TwoFactorAuthenticationProvider::class)->qrCodeUrl(
             config('app.name'),
             $this->{Fortify::username()},
-            $this->twoFactorSetupKey()
+            decrypt($this->two_factor_secret)
         );
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm, does this really accomplish enough to warrant inclusion in core? You can just decrypt this attribute yourself, yes?

jrd-lewis:
> Hmm, does this really accomplish enough to warrant inclusion in core? You can just decrypt this attribute yourself, yes?

You could almost say the same thing about the ```recoveryCodes()``` method in this same trait. It too puts the field in an end-user friendly format, though I don't believe I've seen it used anywhere in the core.

PR #458, opened 2023-04-17
Add ability to override routes with custom paths

stephenglass:
# Backgroud
Fortify does not allow to specify custom paths for its built-in routes. E.g. if you want to change the endpoint name of the built-in `/register` route to `/signup`, this is the current solution:
1. Call `Fortify::ignoreRoutes()` from `register()` in `FortifyServiceProvider.php`
2. Copy all routes from Fortify `routes.php` into your applications `web.php`.
3. Modify the endpoint names as needed.

For E-commerce in particular, you may not want a route to be named "register" as it may have some unwanted connotation depending on your product. 

However, there are a few draw-backs of doing this solution.
a.) You lose some of the functionality that Fortify provides by auto-registering routes based on `Feature` flags and others. As your app scales and requirements change, you may finding yourself going back to implement other routes.
b.) Any future updates of Fortify `routes.php` will need to be manually propagated to your app code even after updating package from composer.

# Proposed Solution
Allow route paths to be overridden via the `fortify.php` config file. Fortify `routes.php` will use the custom name in the config file if provided, otherwise using the default value.

Please see PR for implementation and please let me know if there's any feedback or changes that anyone would like to see!


taylorotwell:
This PR doesn't specify custom names - it specifies custom paths. Is that what you meant?

sglass520:
> This PR doesn't specify custom names - it specifies custom paths. Is that what you meant?

Yes, you are right. I've corrected the PR to state custom paths instead of custom names.

aurac:
Hello, thank you for the feature !

Shouldn't we have some documentation about this ? I struggled a little bit to find that it was possible. In fact I had to find this issue

PR #461, opened 2023-04-20
remove limiter from the registration view route

akunbeben:
I notice that #460 is adding a rate limiter for registration. It shouldn't be applied to the registration view route, in many cases it might we as developers need to refresh the page multiple times for any reason. 

For example, I ran `php artisan optimize:clear` and vite trigger the reload many times here and it will reach the throttle limit.

<details><summary>Details</summary>
<p>

https://user-images.githubusercontent.com/46495960/233508826-1752c0bf-0444-416f-bb34-91f4344bfd4b.mp4

</p>
</details>

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -70,10 +70,7 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
     if (Features::enabled(Features::registration())) {
         if ($enableViews) {
             Route::get(RoutePath::for('register', '/register'), [RegisteredUserController::class, 'create'])
-                ->middleware(array_filter([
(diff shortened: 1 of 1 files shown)


driesvints:
But this is off by default? How is this affecting you?

taylorotwell:
Agree with @driesvints - it's off by default?

trbsi:
I would have to agree with [akunbeben](https://github.com/akunbeben). Rate limiter should not be on a view route. Let's say I want to limit registration 5 per day. If person opens view route 5 times (just opens without actually submitting the form), a person will start getting 429 error

PR #462, opened 2023-04-21
Add missing limiters to default fortify config

stephenglass:
The `config/fortify.php` has all config options listed initialized to null. However, the `limiters` array is missing default assignments for keys `two-factor`, `verification`, `registration` (#460). A user may not know there is rate limiting capabilities for these routes solely by looking at the config unless additionally looked at `routes.php`.

This PR adds default initialization for these limiters in `config/fortify.php` file.

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -15,6 +15,9 @@ return [
     'domain' => null,
     'limiters' => [
         'login' => null,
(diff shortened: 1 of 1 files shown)


taylorotwell:
Tests fail.

Issue #468, opened 2023-05-21
Accept pull request for adding named route for two-factor auth challenge?


OliverGrimsley:
Using Ziggy to publish routes.  Ziggy only pushes named routes to the front end.  There is no named route for POST two-factor challenge.  That code in Fortify is currently:

```php
Route::post(RoutePath::for('two-factor.login', '/two-factor-challenge'), [TwoFactorAuthenticatedSessionController::class, 'store'])
            ->middleware(array_filter([
                'guest:'.config('fortify.guard'),
                $twoFactorLimiter ? 'throttle:'.$twoFactorLimiter : null,
            ]));
```
As a work around, I have added the following to my local routes file:

```php
Route::post(RoutePath::for('two-factor.challenge', '/two-factor-challenge'), [TwoFactorAuthenticatedSessionController::class, 'store'])
    ->middleware(array_filter([
        'guest:' . config('fortify.guard'),
        $twoFactorLimiter ? 'throttle:' . $twoFactorLimiter : null,
    ]))->name('two-factor.challenge');
```

Was going to send a pull request to add this to Fortify, BUT . . . I am not sure if this would break anything for other users to change `RoutePath::for('two-factor.login'`... to `RoutePath::for('two-factor.challenge'...`?  I did not think so because (a) in current Fortify, this is not a NAMED route; and (b) the actual endpoint URL has not been changed.

Happy to send a pull request if this is ok.

nunomaduro:
You can send a pull request, so Taylor Otwell can have the opportunity to understand your needs, and evaluate if your pull request will be breaking for existing users or not.

PR #470, opened 2023-05-25
Update route to name POST two factor challenge

OliverGrimsley:
This pull request changes the POST route for two factor challenge to provide a name for it.  This route was previously un-named.  No change to the actual endpoint is made.  The purpose of this is to allow [Ziggy](https://github.com/tighten/ziggy) to publish this route to the front end.  See issue #468 (current workaround).  I do not think this would break any application as the route was previously un-named, and hence to use it, the actual endpoint would have had to have been referenced.

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -130,11 +130,11 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
                 ->name('two-factor.login');
         }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
It has the same path as the login route, so you can use that name.

haltsir:
I don't understand why this was so swiftly closed since it's a bug. If views are not enabled, named routes are not published.

OliverGrimsley:
My original pull was wrong and so it was properly closed - I'd be happy to submit the above code, which is correct, and would fix this.

haltsir:
If you could submit a fix, that would be very helpful!

PR #471, opened 2023-05-26
Removed direct invocation of `toResponse()` method

xurshudyan:
The `toResponse()` method is called directly on the `FailedTwoFactorLoginResponse` class. However, since `FailedTwoFactorLoginResponse` implements the `Responsable` interface, the `toResponse()` method is automatically called when the response needs to be converted to an HTTP response.


taylorotwell:
If this is fixing an actual user-land bug that breaks applications please re-submit and describe how to recreate it. Thanks.

xurshudyan:
@taylorotwell How could this change potentially break the application's functionality?

PR #474, opened 2023-06-19
This adds the route name two-factor.login to the POST route for /two-factor-challenge when views are not enabled

OliverGrimsley:
This issue is discussed in #470 (my first attempt was incorrect).  This one adds an else statement to the `if ($enableViews)` logic for Two Factor routes.  Under the current route file, if views are _not enabled,_ the POST route for `/two-factor-challenge` (which is named `two-factor.login` if view _ARE_ enabled) is not named, and hence never published if you are using Ziggy.  This just adds the POST route in the else portion, and names it the same thing so ZIggy can pick it up.  

This will not break anyone's code - if views were enabled, the POST route was already named (and this pull request does not change anything for these users); if views were not enabled, a user would not have a named route to use in ZIggy - and would have had to have used a workaround.

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -126,15 +126,21 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
     if (Features::enabled(Features::twoFactorAuthentication())) {
         if ($enableViews) {
             Route::get(RoutePath::for('two-factor.login', '/two-factor-challenge'), [TwoFactorAuthenticatedSessionController::class, 'create'])
-                ->middleware(['guest:'.config('fortify.guard')])
-                ->name('two-factor.login');
-        }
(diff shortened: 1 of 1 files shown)


routes/routes.php
```diff
@@ -126,16 +126,22 @@
     if (Features::enabled(Features::twoFactorAuthentication())) {
         if ($enableViews) {
             Route::get(RoutePath::for('two-factor.login', '/two-factor-challenge'), [TwoFactorAuthenticatedSessionController::class, 'create'])
-                ->middleware(['guest:'.config('fortify.guard')])
-                ->name('two-factor.login');
+            ->middleware(['guest:' . config('fortify.guard')])
+            ->name('two-factor.login');
```
driesvints, line 130: Please undo these indentation changes.

PR #481, opened 2023-08-07
Ensure all test methods are camelCased + add .idea folder to gitignore

umairparacha00:
While reviewing Laravel code, I noticed that all the tests were written in camel case. However, upon examining the Fortify codebase, I found that it used snake case. So, I made the necessary updates and I think camel case is more readable than snake case and the majority of the languages use the camel case.

This is the same as this [PR](https://github.com/laravel/framework/pull/29508) by @lucasmichot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
+/.idea
--- a/tests/AuthenticatedSessionControllerTest.php
+++ b/tests/AuthenticatedSessionControllerTest.php
@@ -20,7 +20,7 @@ use PragmaRX\Google2FA\Google2FA;
 
 class AuthenticatedSessionControllerTest extends OrchestraTestCase
 {
-    public function test_the_login_view_is_returned()
+    public function testTheLoginViewIsReturned()
     {
         $this->mock(LoginViewResponse::class)
                 ->shouldReceive('toResponse')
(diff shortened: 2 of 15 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
+/.idea
```
nunomaduro, line 5: We don't have this change - on any of our repositories.

nunomaduro (changes requested):

lucasmichot:
Hey @umairparacha00 

You will also notice that this PR https://github.com/laravel/framework/pull/39241 was refused, so I would just not bother going on with this change.

I would instead try to enforce the [`php_unit_method_casing`  rule](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/846dc6e63ec41438629e97a280075922010ce5a3/doc/rules/php_unit/php_unit_method_casing.rst#L4) within pint, for the [Laravel](https://github.com/laravel/pint/blob/main/resources/presets/laravel.php) preset.

.gitignore
umairparacha00: I saw the .idea was not added in the gitignore so added it.

if you meant something else then plz explain.

umairparacha00:
@lucasmichot 
This [PR](https://github.com/laravel/framework/pull/39241) was not in my knowledge. 
Thanks for the clarification.

PR #482, opened 2023-08-10
Password reset - allow success response on invalid user

Bosbaba:
### Small security feature:

Adding a config flag to allow for responses to appear as successful when invalid emails are submitted during password resets, preventing intruders from fishing for valid email addresses

When used in combination with laravel, I would suggest just updating the reset success message to be more verbose in `resources/lang/en/passwords.php` : `sent`

Something like this:
`An email has been sent to the address provided, if it exists in our system. Please check your inbox for further instructions on how to reset your account. If you don't receive an email within a few minutes, please ensure that the email address entered is correct and try again. For security reasons, we do not confirm the existence of email addresses in our system.`


taylorotwell:
Even after this PR they could just attempt to register and see if the email is already taken?

PR #484, opened 2023-08-19
Post-registration "no login" option

curiousyigit:
This PR adds the ability to register without login. This seems to be a pretty common need among users. The alternative being overriding the RegisteredController & route, this is a much more elegant solution.

Example usage:
```
// in config/fortify.php

    /*
    |--------------------------------------------------------------------------
    | Features
    |--------------------------------------------------------------------------
    |
    | Some of the Fortify features are optional. You may disable the features
    | by removing them from this array. You're free to only remove some of
    | these features or you can even remove all of these if you need to.
    |
    */

    'features' => [
        Features::registration([
            'login' => false,
        ]),
    ],

];
```

### DIFF
--- a/src/Features.php
+++ b/src/Features.php
@@ -74,10 +74,14 @@ class Features
     /**
      * Enable the registration feature.
      *
+     * @param  array  $options
      * @return string
      */
-    public static function registration()
+    public static function registration(array $options = [])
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #487, opened 2023-09-02
Update logout to invalidate and regenerate session only if session is present

karmendra:
Added check in logout to invalidate and regenerate session only if request has session. This will allow logout to work when 'api' middleware is used with fortify.


driesvints:
As said on the issue, this isn't a real problem since an API is stateless and therefor should never use this method.

PR #492, opened 2023-09-27
fix: TwoFactorDisabledResponse and TwoFactorEnabledResponse should implement their respective interfaces

BJTheCod3r:
- This correction addresses the incorrect implementation of interfaces by TwoFactorDisabledResponse and TwoFactorEnabledResponse. Both classes currently implement the TwoFactorLoginResponse interface, but this fix introduces the appropriate implementations for their respective interfaces.


taylorotwell:
I think this would be a breaking change on a patch release.

PR #496, opened 2023-10-06
Allow specifying a prefix for route names

moisish:
The PR introduces a way of setting a prefix on the name of the routes. 

The reason of the PR is if you are building an admin panel and you want to use fortify, it makes sense all the route names to be 'admin.' such as 'admin.login', 'admin.logout'. Especially when you have a frontend where it has it's own Authentication and you need to separate the routes

A new option has been added the config file which by default it keeps the existing behaviour.

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -13,6 +13,7 @@ return [
     'home' => '/home',
     'prefix' => '',
     'domain' => null,
+    'name_prefix' => null,
     'lowercase_usernames' => false,
     'limiters' => [
         'login' => null,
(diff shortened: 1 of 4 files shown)


moisish:
@driesvints The changes on this PR affects Jetstream. How can I handle this case?

moisish:
Should I get the jetsteam changes done as well before I marked this ready?

driesvints:
@moisish I don't know. I'd just mark this as ready so Taylor can give his opinion.

taylorotwell:
/

moisish:
@taylorotwell anything i can do  to get this merged?

PR #497, opened 2023-10-11
Use trans option in failed password confirmation response

gustavosobrinho01:
Currently, when using Laravel's translation feature in the password confirmation process, if you enter an invalid password, the translation is not applied, and it remains as an English message, rather than in the locale defined in the config/app.php file.

`Message displayed:`

Captura de Tela 2023-10-10 às 22 23 10

`Locale applied:`

Captura de Tela 2023-10-10 às 22 24 05


taylorotwell:
Put the translation in a JSON translation file.

PR #499, opened 2023-10-11
Name all routes in Fortify

### LINKED ISSUE #498: Name all routes for Ziggy support when enableViews is false
### Fortify Version

1.16

### Laravel Version

10.10

### PHP Version

8.1

### Database Driver & Version

_No response_

### Description

The routes defined in https://github.com/laravel/fortify/blob/master/routes/routes.php don't always have names. The routes leading to views are always named when `enableViews` is true. But routes leading to backend actions used when enableViews is false are not always named.

Names can still be useful when using Ziggy to link the frontend to the Laravel route definitions. For example you can call logout by name, but not login. Here's the Fortify route definitions:

```
    Route::post('/login', [AuthenticatedSessionController::class, 'store'])
        ->middleware(array_filter([
            'guest:'.config('fortify.guard'),
            $limiter ? 'throttle:'.$limiter : null,
        ]));

    Route::post('/logout', [AuthenticatedSessionController::class, 'destroy'])
        ->name('logout');
```

This is similar to #468 and #470. The latter led to a draft PR #474 However imho that looks a bit complex because it attempts to assign the name `two-factor.login` to either the create action (if `$enableViews` is true) or the store action (if `$enableViews` is false). It seems to me an easier solution would be to leave all existing names as they are and just add new names to the unamed routes. For example in the case of `two-factor.login`,

```php
        if ($enableViews) {
            Route::get('/two-factor-challenge', [TwoFactorAuthenticatedSessionController::class, 'create'])
                ->middleware(['guest:'.config('fortify.guard')])
                ->name('two-factor.login'); // Leave this as-is
        }

        Route::post('/two-factor-challenge', [TwoFactorAuthenticatedSessionController::class, 'store'])
            ->middleware(array_filter([
                'guest:'.config('fortify.guard'),
                $twoFactorLimiter ? 'throttle:'.$twoFactorLimiter : null,
            ]))
            ->name('two-factor.login.store'); // Just add a new name for this
```

This would not be a breaking change and would support those using Ziggy. The ultimate goal being that once using Ziggy it is more consistent to use route names throughout the frontend, rather than mixing route names and paths.

### Steps To Reproduce

Install Laravel and Fortify. Set `'views' => false,` in `config/fortify/php`/ Run `php artisan route:list`. Examine Fortify routes. Some are named, some not. For example, 

```
 POST            login .......................................................... Laravel\Fortify › AuthenticatedSessionController@store
 POST            logout .............................................. logout › Laravel\Fortify › AuthenticatedSessionController@destroy
```

sfreytag:
This Pull Request proposed added names to all Fortify routes such that they can be consistently used by name with Ziggy in a frontend when Fortify's enableViews is set to false.

It uses new names for the unnamed routes such that there is no breaking change and no competition for eg `login`. Rather than assigning the name `login` to the create action when `enableViews` is true and to the store action when `enableViews` is false it seems clearer to just use two names, `login` and `login.store`.

This suffix `.store` is used consistently throughout for the new names because they're all store actions.

#498 provides more explanation.

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -36,7 +36,8 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
         ->middleware(array_filter([
             'guest:'.config('fortify.guard'),
             $limiter ? 'throttle:'.$limiter : null,
-        ]));
+        ]))
+        ->name('login.store');
 
     Route::post('/logout', [AuthenticatedSessionController::class, 'destroy'])
         ->name('logout');
@@ -71,7 +72,8 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
         }
 
         Route::post('/register', [RegisteredUserController::class, 'store'])
-            ->middleware(['guest:'.config('fortify.guard')]);
+            ->middleware(['guest:'.config('fortify.guard')])
+            ->name('login.store');
     }
 
     // Email Verification...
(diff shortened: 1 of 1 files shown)


taylorotwell:
What are you actually prevented from doing given the current names?

sfreytag:
> What are you actually prevented from doing given the current names?

Using Ziggy and named routes throughout the frontend. EG I can call `axios.post(route('logout'))` but not the equivalent for login; I have to do `axios.post('/login')`.

PR #506, opened 2023-11-15
Uses PHP Native Type Declarations

slvler:


driesvints:
Thanks but right now we don't have plans for this.

PR #507, opened 2023-11-27
[develop] Adds L11 support

nunomaduro:
This pull request adds L11 support to fortify.

Also, drops:
- PHP 7.3, 7.4, 8.0.
- Laravel 8 and 9.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "mockery/mockery": "^1.0",
         "orchestra/testbench": "^8.16|^9.0",
         "phpstan/phpstan": "^1.10",
-        "phpunit/phpunit": "^10.0"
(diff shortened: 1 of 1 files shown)


crynobone (approved):

stubs/fortify.php
```diff
@@ -74,7 +73,7 @@
     |
     */
 
-    'home' => RouteServiceProvider::HOME,
+    'home' => '/home',
```
taylorotwell, line 76: I think we should unify on `/dashboard` here, which is what Jetstream uses.

nunomaduro: @taylorotwell I apply that change via Jetstream's install command.

PR #510, opened 2023-12-11
Use the illuminate password rule

ricklambrechts:
I think it would be nice if we use the Password rule that is available so we do not have seperate classes with different functionalities.

The [`Illuminate\Validation\Rules\Password`](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Validation/Rules/Password.php) has more functionality than the Fortify implementation. For example this rule has mixed case and compromised check available.


driesvints:
Although I'm not opposed, this is a breaking change for a minor release. You can't simply delete a class from a minor release.

driesvints:
Try the master branch maybe?

driesvints:
Or maybe better: re-send this without deleting the class and simply add a `@deprecated` annotation to it on top.

ricklambrechts:
Thanks dries!

PR #511, opened 2023-12-11
Deprecate the password rule and use illuminate password rule

ricklambrechts:
Created a new PR with the feedback of Dries onto https://github.com/laravel/fortify/pull/510

This will deprecate the Fortify password rule and use the `Illuminate\Validation\Rules\Password` rule.

The `Illuminate\Validation\Rules\Password` rule has more functionality than the Fortify rule. For example this rule has mixed case and compromised check available.

### DIFF
--- a/src/Rules/Password.php
+++ b/src/Rules/Password.php
@@ -5,6 +5,9 @@ namespace Laravel\Fortify\Rules;
 use Illuminate\Contracts\Validation\Rule;
 use Illuminate\Support\Str;
 
+/**
+ * @deprecated Use \Illuminate\Validation\Rules\Password instead.
+ */
 class Password implements Rule
 {
     /**
--- a/stubs/PasswordValidationRules.php
+++ b/stubs/PasswordValidationRules.php
@@ -2,7 +2,7 @@
 
 namespace App\Actions\Fortify;
 
-use Laravel\Fortify\Rules\Password;
+use Illuminate\Validation\Rules\Password;
 
 trait PasswordValidationRules
 {
@@ -13,6 +13,6 @@ trait PasswordValidationRules
      */
     protected function passwordRules(): array
     {
-        return ['required', 'string', new Password, 'confirmed'];
+        return ['required', 'string', Password::default(), 'confirmed'];
     }
 }


onlime:
Hi @ricklambrechts I very much favor deprecating any duplicate code and I was always wondering why `Laravel\Fortify\Rules\Password` was ever created. The problem though is, that `Illuminate\Validation\Rules\Password` offers more features like the compromised check, but provides less detailed validation messages. So, there is no feature parity.

Look at this example, when validating password "password"  with below rules:

`Laravel\Fortify\Rules\Password`:
> `The password must be at least 10 characters and contain at least one uppercase character and one number.`

vs. `Illuminate\Validation\Rules\Password`:
> `The string must be at least 10 characters.`

Personally, I was always using a combination (composition rule) of both like this, for exactly that reason:

```php
use Illuminate\Validation\Rules\Password;
use Laravel\Fortify\Rules\Password as FortifyPassword;

class PasswordStrict extends CompositeRule
{
    protected function rules(): array
    {
        return [
            'string',
            'max:150',
            // Fortify Password rule: provides more specific validation error messages
            // same as: Password::min(10)->mixedCase()->numbers()
            (new FortifyPassword)
                ->length(10)
                ->requireUppercase()
                ->requireNumeric(),
            // Illuminate Password rule: simpler individual error messages, but provides uncompromised()
            // see Password::defaults() in AppServiceProvider::boot() method
            Password::defaults(),
            new ZxcvbnRule,
            'confirmed',
        ];
    }
```

For better UX, I totally need such compound validation messages like in Fortify's `Password` rule. But I understand it's quite tricky to build them (see [Password.php#L85-L139](https://github.com/laravel/fortify/blob/1.x/src/Rules/Password.php#L85-L139)).

How should we proceed? PR that feature from Fortify into Illuminate `Password`?

driesvints:
Hi @onlime. Sorry about that. We've decided to leave things be for now, sorry.

onlime:
> Hi @onlime. Sorry about that. We've decided to leave things be for now, sorry.

What does that mean, removing the `@deprecated` again?

driesvints:
No, we're not going to make additional changes sorry.

PR #518, opened 2024-01-31
Don't overwrite an already two factor secret unless force = true

danmatthews:
Add a `force` parameter that only overwrites existing two factor secret keys and recovery codes in the database if present when calling the enable two factor authentication action.

## Why

If the secret has already been set, the QR code may have already been shown to the end user, and this endpoint is erroneously called again - through a page refresh or some JS issue, the secret is overridden, rendering the code the end user just scanned useless without informing them that anything has changed.

## Mitigation

When calling the `/user/two-factor-authentication` endpoint, the code now checks to see if the `two_factor_secret` is not `null` and assumes that it shouldn't overwrite it to prevent situations like this.

To force it to overwrite, you can pass the `force` parameter - which will fill out the two factor columns with new values as it did previously.

The note here is to focus on **intent** - a piece of code should only overwrite this through the intent of the user (maybe clicking a button that refreshes the QR code before scanning it) or through some app-level action like resetting a user's twoFA status.

### DIFF
--- a/src/Actions/EnableTwoFactorAuthentication.php
+++ b/src/Actions/EnableTwoFactorAuthentication.php
@@ -31,17 +31,22 @@ class EnableTwoFactorAuthentication
      * Enable two factor authentication for the user.
      *
      * @param  mixed  $user
+     * @param  bool   Should we overwrite the currently set secret if one exists.
      * @return void
      */
-    public function __invoke($user)
+    public function __invoke($user, $force = false)
     {
-        $user->forceFill([
-            'two_factor_secret' => encrypt($this->provider->generateSecretKey()),
-            'two_factor_recovery_codes' => encrypt(json_encode(Collection::times(8, function () {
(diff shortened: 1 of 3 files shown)


driesvints:
Can you please tell us more about your use case?

danmatthews:
@taylorotwell sure - upon clicking a button to open a modal to enable two factor auth in your account, the `onMount()` would send a request to the 'enable' endpoint and upon receiving a `200` response, would use Inertia's partial reloads to load the QR code SVG.

Because we have URL based persistent modals, refreshing the page after scanning the QR code would call the endpoint again, resulting in the secret being overridden, but the user would have already scanned it.

It took us a LONG time to track this behaviour down as it was such an edge case as a result of user behaviour - but it meant that people would be locked out.

We were surprised to find that multiple calls to the enable end point didn't check to see if it had already been 'activated', that's how we resulted in this PR.

Issue #339, opened 2022-01-08
Fortify::authenticateUsing called twice


oulfr:
- Fortify Version: 1.8
- Laravel Version: 8.65
- PHP Version: 8.0
- Database Driver & Version:

### Description:
The **Fortify::authenticateUsing** method is triggered twice for each request.

### Steps To Reproduce:
In FortifyServiceProvider create an index:
     `protected  $i=1;`

In boot method add tis function:

        Fortify::authenticateUsing(function (Request $request) {
            $this->i++;

            if( $this->i > 2){
                dd('hmm', $this->i);
            }
            $user = User::where('email', $request->email)->first();
            if ($user && Hash::check($request->password, $user->password)) {
                return $user;
            }
        });

        Fortify::loginView(function () {
            return view('user::auth.login');
        });

You will get: 
            "hmm"
            3

driesvints:
I'm not really sure what the issue is. If you can post much more clear description what is amis and how it blocks you with clear steps to reproduce then we can have another look. Thanks

oulfr:
The issue is the **Fortify::authenticateUsing** method is triggered twice.
It should be called one time for each request

driesvints:
I'll try it on Monday

driesvints:
Bit swamped with Laravel 9 stuff right now so I'm delaying this until I find some time.

driesvints:
I indeed get the same thing but since there's no actual bug and nothing's broken I'm going to close this. If you can help us optimize this we'd appreciate a PR. Thanks

ricardogobbosouza:
Hi @oulfr 
It actually happens...see https://github.com/laravel/fortify/pull/304

driesvints:
@ricardogobbosouza I didn't know about that PR anymore. Looks like this is a wont fix, sorry.

ricardogobbosouza:
The authentication method is called twice when the `twoFactorAuthentication` feature is enabled

https://github.com/laravel/fortify/blob/1.x/src/Http/Controllers/AuthenticatedSessionController.php#L86-L87

oulfr:
this bug is blocked if we want to add google recaptcha validation to the authentication form, because the google recaptcha api only authorizes the first request and it considers the second call as a bot and never the validation will pass.
to resolve the issue we have implemented something like this:

              class FortifyServiceProvider extends ServiceProvider
              {
                  protected $isFirstCall = false;
	              
	              public function boot()
                  {
                      
[...]

PR #532, opened 2024-04-23
Fix authenticateUsing called twice

### LINKED ISSUE #530: authenticateUsing called twice
### Fortify Version

1.21.1

### Laravel Version

11.4.0

### PHP Version

8.3.3

### Database Driver & Version

n/a

### Description

The Fortify::authenticateUsing method is triggered twice for each request.

### Steps To Reproduce

This is the same issue as this previous issue: https://github.com/laravel/fortify/issues/339

Considering this is in fact a bug, and your Issues queue is empty, would you please considering fixing this as it is a bug and a known problem.

Perhaps having a discussion about it at least without just closing the issue would be helpful, at least for someone/anyone to fix it?

webdevnerdstuff:
This is to fix the issue of `authenticateUsing` getting called twice.

Fixes: [Issue 530](https://github.com/laravel/fortify/issues/530) & [Issue 339](https://github.com/laravel/fortify/issues/339)

### DIFF
--- a/src/Actions/AttemptToAuthenticate.php
+++ b/src/Actions/AttemptToAuthenticate.php
@@ -69,7 +69,7 @@ class AttemptToAuthenticate
      */
     protected function handleUsingCustomCallback($request, $next)
     {
-        $user = call_user_func(Fortify::$authenticateUsingCallback, $request);
+        $user = $request['authUser'] ?? call_user_func(Fortify::$authenticateUsingCallback, $request);
 
         if (! $user) {
             $this->fireFailedEvent($request);
--- a/src/Actions/AttemptToAuthenticate.php
+++ b/src/Actions/AttemptToAuthenticate.php
@@ -69,7 +69,7 @@ class AttemptToAuthenticate
      */
     protected function handleUsingCustomCallback($request, $next)
     {
-        $user = $request['authUser'] ?? call_user_func(Fortify::$authenticateUsingCallback, $request);
+        $user = $request->get('user') ?? call_user_func(Fortify::$authenticateUsingCallback, $request);
 
         if (! $user) {
             $this->fireFailedEvent($request);
--- a/src/Actions/RedirectIfTwoFactorAuthenticatable.php
+++ b/src/Actions/RedirectIfTwoFactorAuthenticatable.php
@@ -48,7 +48,7 @@ class RedirectIfTwoFactorAuthenticatable
      */
     public function handle($request, $next)
     {
-        $user = $this->validateCredentials($request);
+        $user = $request['authUser'] ?? $this->validateCredentials($request);
 
         if (Fortify::confirmsTwoFactorAuthentication()) {
             if (optional($user)->two_factor_secret &&
--- a/src/Actions/RedirectIfTwoFactorAuthenticatable.php
+++ b/src/Actions/RedirectIfTwoFactorAuthenticatable.php
@@ -39,9 +39,6 @@ class RedirectIfTwoFactorAuthenticatable
         $this->limiter = $limiter;
     }
 
-
-    public $authUser;
-
     /**
      * Handle the incoming request.
      *
--- a/src/Actions/RedirectIfTwoFactorAuthenticatable.php
+++ b/src/Actions/RedirectIfTwoFactorAuthenticatable.php
@@ -39,6 +39,9 @@ class RedirectIfTwoFactorAuthenticatable
         $this->limiter = $limiter;
     }
 
+
+    public $authUser;
+
     /**
      * Handle the incoming request.
      *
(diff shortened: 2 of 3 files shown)


taylorotwell:

webdevnerdstuff:
@taylorotwell this is a bug as explained in both the Issues mentioned.  One being is it's causing requests to be considered a bot, and problems with rechapcha, which is not good. The problem is that Fortify is causing to validate the login multiple times, which is not what it should be doing and I'm pretty sure not an intended behavior, and is also inefficient by repeating an unnecessary one right after the other.  This s a bug, not just a inconvenience. It has multiple reasons it will cause problems for people. You can read more about them here: [Issue 339](https://github.com/laravel/fortify/issues/339)

The code I added is literally 3 lines of very short code, plus adjusting another line slightly, so it is not something I would consider as a big maintenance factor for you, and the code added is so basic it only takes a few seconds to understand. It's also definitely not worth creating an entire package for considering. It also fixes a major problem for people who it affects. All the new code is doing is passing along the validated user so the next step does not cause it to go right back to the same place to literally run it all over again to get the same exact data.

I have also spent many hours debugging what the issue was in the first place, then trying to find a solution, as well as trying to figure out work arounds, then searching the issues only to find it's a bug that's been round for a year. Then spent many more hours trying to fix this issue being completely unfamiliar with your code base and having to learn what everything was doing and the flow because your team will not even consider fixing it for some unknown reason. This could also force my team to have to throw away weeks worth of work to find a new solution. Which is a bit frustrating to say the least.

It's a very simple change, yet solves a major problem for people, so I really hope you considering accepting this very minimal change.

driesvints:
@webdevnerdstuff looks like this is an absolute no-fix for us sorry. We just don't want to risk a breaking change for others.

webdevnerdstuff:
> @webdevnerdstuff looks like this is an absolute no-fix for us sorry. We just don't want to risk a breaking change for others.

Here's an idea. Next time don't ask other people to waste their time fixing bugs in your code because you either don't care or are too lazy to do it yourself and have had well over a year to do it, only to give them ridiculous (and invalid) reasons as to why you won't except a PR that fixes your bugs. You don't even discuss it or try to have an intelligent discussion about it.

Yes, they are ridiculous excuses because not only did you not look at the code, you didn't even give a valid reason or explanation, or possible solutions, you just close it and give excuses, thinking others are too stupid to know they are ridiculous excuses. All it does is waste peoples times (including yours), annoys and insults them, and show to others how you clearly do not care about the quality of your code or the people who use it. It's unprofessional and gives OSS development a bad name.

I also don't need you to respond with some other excuses or empty apologies. I needed to say my peace because I really do not appreciate other people blatantly wasting my time. So at the very least give me the professional curtesy of not responding so I don't feel the need to respond again, but next time not being so professional about it.

siubie:
Finally i know why my recaptcha complained about duplicate request when i validate user captcha, I will be waiting for this fix.

I use fortify and i can confirm if AuthenticateUsing being called twice thats why the recaptcha api respond with timeout-or-duplicate.

Thanks @webdevnerdstuff

reniar1:
Is there any fix for this yet?

dan-rprado:
Deactivating `Features::twoFactorAuthentication ` fixes the issue, but I still don't think that's a proper solution.

Chomiciak:
@webdevnerdstuff The fix proposed is indeed not mergeable, as it does save an attribute under a name, that may be used by the user's implementation, thus can be a breaking change. 

However, not fixing such a simple problem in place, when it is fully discovered, doesn't hint any good as of development quality in Laravel team...

kratos-digital:
In FortifyServiceProvider file, under boot method, using code directly from docs and adding my own logging action:

```
public function boot(): void
{
    Fortify::authenticateUsing(function (Request $request) {
            $user = User::where('email', $request->email)->first();

            if ($user &&
                Hash::check($request->password, $user->password)) {
                // This is my custom code, while everything else is from docs
                activity()
                    ->performedOn($user)
                    ->causedBy($user)
                    ->log(ActivityLog::LOGGEDIN);

                return $user;
            }
    });
}
```

Log is saved TWICE in database. This is bug. Somewhere. Obviously. As reported by original poster. 2 months ago.
I found same issue reported 3 years ago lol.

Deactivating 2FA is not a solution, because it strips you of the feature.

Is there a way around or "quick" fix without changing the source code while keeping 2FA? @taylorotwell

tfevan:
I'm facing the same issue. My 'captcha' validation run twice, that's why I cant use captcha properly. Everything works well when I disabled the 2FA. But this is not the Solution :(

putrafajarh:
I have facing same issue to integrate Cloudflare Turnstile with Inertia, 
Temp solution disable twoFactorAuthentication

PR #534, opened 2024-04-25
[1.x] Bacon QR 3.0 support

eshimischi:
https://github.com/Bacon/BaconQrCode/releases/tag/v3.0.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "require": {
         "php": "^8.1",
         "ext-json": "*",
-        "bacon/bacon-qr-code": "^2.0|^3.0",
+        "bacon/bacon-qr-code": "^3.0",
         "illuminate/support": "^10.0|^11.0",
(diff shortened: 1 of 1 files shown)


driesvints:
Tbh, think we can just bump to 3.0 since we're also at a minimum of PHP 8.1

eshimischi:
> Tbh, think we can just bump to 3.0 since we're also at a minimum of PHP 8.1

Yes indeed! I checked my personal project with 3.0, no issues

Jhamnerx:
When updating to this version I have not been able to install
simplesoftwareio/simple-qrcode
because it does not have compatibility with baconqr 3.0

eshimischi:
> simplesoftwareio/simple-qrcode

https://github.com/SimpleSoftwareIO/simple-qrcode/issues/304 as far as adapt a new version of baconqr, push them forward

PR #536, opened 2024-05-08
[1.x] Ensure logout route is authenticated

timacdonald:
It is currently possible to hit the `/logout` route as a guest. This means that the auth system "logs out" a `null` user.

This can occur in the real world when you:

1. Login to an app
2. Open the app in two tabs
3. Logout of one tab
4. Logout of the second tab

The framework will now logout a `null` user.

```php
Event::listen(fn (Logout $event) => assert($event->user === null));
```

I believe this is a good fix because:

1. Guests should not be able to visit the logout route. It is explicitly for authenticated users – just like the login route is explicitly for guests.
2. If you are listening for the logout event you are expecting a user.
3. The event is also typed (docblock) to require a user object.
4. Lastly, `laravel/breeze` does this, so it creates consistency for Jetstream and Fortify.

This issue was raised because 2 users have run into this in Pulse when using Jetstream (https://github.com/laravel/pulse/issues/364) and Laravel UI (https://github.com/laravel/pulse/issues/324). Although we will also handle this better in Pulse, I believe this is a good fix to include in Fortify.

- `laravel/ui` fix: https://github.com/laravel/ui/pull/269

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -42,6 +42,7 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
         ]));
 
     Route::post(RoutePath::for('logout', '/logout'), [AuthenticatedSessionController::class, 'destroy'])
+        ->middleware([config('fortify.auth_middleware', 'auth').':'.config('fortify.guard')])
         ->name('logout');
 
     // Password Reset...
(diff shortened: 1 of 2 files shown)


adiachenko:
This change broke one of our frontend app after attempting at upgrade to new version because it relied on logout having **idempotent** behavior and always returning 204 whether or not you has an active session.

Overall, I think it was an unnecessary change because it doesn't really solve any issues aside from not having to handle null safety on logout event and breaks behavior that some apps may rely on.

driesvints:
Hi @adiachenko. Sorry you got caught with this but it was never intended for this endpoint to non-stateful. We won't be reverting this change.

PR #537, opened 2024-05-11
Add rate limiter for password update

duckarcher:
# Overview

A minor improvement that allows customizing the rate limiting of the `user-password.update` route. It would benefit those who want to rate limit the password update route for some reason (I will use it on my project since I like to be as strict as possible with rate limiting). There is no specific security concern here for the specific route other than the fact that it's nice to allow controlling the rate limiting of routes.

> Note: Change is backwards compatible.

# Testing

Add the line:
```
'password-update' => '1,1'
```
on `limiters` section of `fortify.php` file. Then, hit the password update endpoint twice. You will get a "Too Many Attempts" response for the 2nd request.

> Note: Sorry I didn't add a test, I couldn't set the config inside the test for some reason. I tested it manually by doing the above and messing with the `test_passwords_can_be_updated` test.


taylorotwell:

PR #540, opened 2024-06-04
Remove repeated check that feature is enabled for 2fa confirmation 

czernika:
Minor refactoring

Removed repeated check if feature is enabled (`Feature::enabled()`) as it is present within `optionEnabled` [method](https://github.com/laravel/fortify/blob/1.x/src/Features.php#L25)


taylorotwell:
No plans to change right now if it's not broken.

Issue #543, opened 2024-06-19
Login throttle ternary condition inside `AuthenticatedSessionController` seems incorrect


brnquester:
### Fortify Version

1.21.3

### Laravel Version

10.48.12

### PHP Version

8.2

### Database Driver & Version

MySQL 8.0 Docker container mysql/mysql-server:8.0

### Description

It seems that the ternary condition inside of `AuthenticatedSessionController` is inverted for `EnsureLoginIsNotThrottled` verification.

Location:
https://github.com/laravel/fortify/blob/a725684d17959c4750f3b441ff2e94ecde7793a1/src/Http/Controllers/AuthenticatedSessionController.php#L85-L91

It should like below:
```php
        return (new Pipeline(app()))->send($request)->through(array_filter([
            config('fortify.limiters.login') ? EnsureLoginIsNotThrottled::class : null,
            config('fortify.lowercase_usernames') ? CanonicalizeUsername::class : null,
            Features::enabled(Features::twoFactorAuthentication()) ? RedirectIfTwoFactorAuthenticatable::class : null,
            AttemptToAuthenticate::class,
            PrepareAuthenticatedSession::class,
        ]));
```

### Fortify own lockout handler is needed for different scenarios

- https://github.com/laravel/fortify/issues/336 
- https://github.com/laravel/fortify/issues/312

### Ideal scenario / my proposal
Have both protections by having different throttle configs for the Laravel middleware and Fortify own handler. We can have a smaller throttle rate for Fortify's own handler and a greater one for the middleware just in case of brute force. 

### Steps To Reproduce

1. Make sure to have the limiters defined in `config/fortify.php` as follows:
```
'limiters' => [
        'login' => 5,
        'two-factor' => 5,
    ],
```
2. Remove the throttle middleware from the `Route::post(RoutePath::for('login', '/login')` keeping everything else as is, see below: (Fortify has its own throttle error handlers, it shouldn't need to rely on Laravel defaults throttle middleware)
```
Route::post(RoutePath::for('login', '/login'), [AuthenticatedSessionController::class, 'store'])
        ->middleware(array_filter([
            'guest:'.config('fortify.guard')
        ]));
```
4. Now attempt multiple logins passing the limiter mark, you can try even 20 times. You'll never see the error "Too many login attempts. Please try again in X seconds.".
[...]

Issue #336, opened 2021-12-25
Lockout event is never fired 


alperyazgan:
- Fortify Version: 1.8.6
- Laravel Version: 8.77.1
- PHP Version: 8.1.1

## What is aimed
 Enable a configurable option for the app to use either the throttling or user lockout. 
## What is failing
Throttling is working fine but the lockout event is never fired so never fetched, lockout option could not be setup.  
### Description:

With 2 Different Case Scenarios, during Authentication (with same user, wrong password) lockout event is expected to be fired but never happened;

**First Scenario:**
 config/fortify.php has the following limiters set 
` 'limiters' => [
    'login'      => 'login',
    'two-factor' => 'two-factor',
  ],
`
 And inside the boot method of "_FortifyServiceProvider_" we have _authenticateUsing_ and _authenticateThrough_ methods configured and the _rate limiters_.

```php
public function boot()
  {
    Fortify::viewPrefix('app.auth.');

    Fortify::createUsersUsing(CreateNewUser::class);
    Fortify::updateUserProfileInformationUsing(UpdateUserProfileInformation::class);
    Fortify::updateUserPasswordsUsing(UpdateUserPassword::class);
    Fortify::resetUserPasswordsUsing(ResetUserPassword::class);

    Fortify::authenticateUsing(function (Request $request) {
      $loginUsername = config('fortify.username');

      $user = User::where([
                            [$loginUsername, '=', strtolower($request->input($loginUsername))],
                            ['active', '=', true],
                            ['locked', '=', false],
                          ])->first();
      if ($user && Hash::check($request->password, $user->password)) return $user;
    });

    Fortify::authenticateThrough(function (Request $request) {
      return [
        EnsureLoginIsNotThrottled::class,
        AttemptToAuthenticate::class,
        PrepareAuthenticatedSession::class,
      ];
    });

    RateLimiter::for('login', function (Request $request) {
      $loginUsername = config('fortify.username');

      return Limit::perMinute(config('app-settings.max_login_attempts'))->by($request->$loginUsername . $request->ip());
    });

    RateLimiter::for('two-factor', function (Request $request) {
      return Limit::perMinute(5)->by($request->session()->get('login.id'));
    });
  }
```
[...]

Issue #312, opened 2021-09-06
Throttle middleware does not play nicely with Inertia


bastinald:
I'm trying to use Fortify with Inertia and everything works great except the throttle middleware is not returning a status or error which I can use inside my Vue components. 

Instead, I get this: https://i.imgur.com/yWpXMVp.png

Am I missing something? How can I get the throttle middleware that Fortify uses to give me a `status` or `errors` which I can use in my Vue components? I'd need this for login, verify email, etc.

bastinald:
submitted PR so we can set our own limiter response like we can with `login`: https://github.com/laravel/fortify/pull/313

brnquester:
@bastinald - I understand this issue is old, but the Lockout Fortify own handler that plays nicely with InertiaJS is still present but never reached due to the Laravel middleware. However, with simple steps, I was able to get it working back again. This to me seems like an existing bug. The Laravel middleware should be an option that we enable or not, but never the only option. Any support in my current issue is appreciated.
https://github.com/laravel/fortify/issues/543

PR #544, opened 2024-06-20
Fixed issue with login throttle condition and conflict with Laravel middleware

brnquester:
# Overview 

This change will fix an issue where the `EnsureLoginIsNotThrottled` verification is never reached if `fortify.limiters.login` has any value. The current ternary condition logic is inverted. 

Also, it solves the conflict of the Laravel throttle middleware cutting the flow to Fortify throttle's own handlers by adding more granular control over the rate limits for Fortify, Laravel middleware, and the decay in seconds. 

Finally, this solution will keep both throttle protections of Fortify's own handler and Laravel default middleware. If the configs proposed here are followed, the Fortify `EnsureLoginIsNotThrottled` will be triggered first returning a friendly message `Too many requests. Please try again in X seconds.` which plays nicely with InertiaJS and if the requests continue replicating a brute force behaviour, then Laravel kicks in.

## Steps to reproduce without this fix

```
'limiters' => [
        'login' => 5,
        'two-factor' => 5,
    ],
```
```
        ->middleware(array_filter([
        ]));
```

## Relevant issues found
- https://github.com/laravel/fortify/issues/543

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -31,14 +31,14 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
             ->name('login');
     }
 
-    $limiter = config('fortify.limiters.login');
-    $twoFactorLimiter = config('fortify.limiters.two-factor');
+    $limiterMiddleware = config('fortify.limiters.login-middleware');
+    $twoFactorLimiterMiddleware = config('fortify.limiters.two-factor-middleware');
     $verificationLimiter = config('fortify.limiters.verification', '6,1');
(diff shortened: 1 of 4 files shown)


driesvints:
@brnquester seems tests fail here. Could you resend this one with passing tests? Thanks!

brnquester:
> @brnquester seems tests fail here. Could you resend this one with passing tests? Thanks!

Ok. I'll come back later then, forgot about the tests.

PR #545, opened 2024-06-20
Fixes Login throttle ternary condition inside `AuthenticatedSessionController`

### LINKED ISSUE #543: Login throttle ternary condition inside `AuthenticatedSessionController` seems incorrect

crynobone:
fixes #543

### DIFF
--- a/src/Http/Controllers/AuthenticatedSessionController.php
+++ b/src/Http/Controllers/AuthenticatedSessionController.php
@@ -83,7 +83,7 @@ class AuthenticatedSessionController extends Controller
         }
 
         return (new Pipeline(app()))->send($request)->through(array_filter([
-            config('fortify.limiters.login') ? null : EnsureLoginIsNotThrottled::class,
+            config('fortify.limiters.login') ? EnsureLoginIsNotThrottled::class : null,
             config('fortify.lowercase_usernames') ? CanonicalizeUsername::class : null,
             Features::enabled(Features::twoFactorAuthentication()) ? RedirectIfTwoFactorAuthenticatable::class : null,
             AttemptToAuthenticate::class,
--- a/tests/AuthenticatedSessionControllerTest.php
+++ b/tests/AuthenticatedSessionControllerTest.php
@@ -17,6 +17,7 @@ use Laravel\Fortify\Features;
 use Laravel\Fortify\LoginRateLimiter;
 use Laravel\Fortify\TwoFactorAuthenticatable;
 use Mockery;
+use Orchestra\Testbench\Attributes\WithConfig;
 use Orchestra\Testbench\Attributes\WithMigration;
 use PragmaRX\Google2FA\Google2FA;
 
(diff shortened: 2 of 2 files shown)


taylorotwell:
The code is correct. 

If the limiter config is set we don't use Fortify's built-in Limiter `EnsureLoginIsNotThrottled` - we instead use the limiter the user defines which is assigned to the route in Fortify's route file.

brnquester:
> The code is correct.
> 
> If the limiter config is set we don't use Fortify's built-in Limiter `EnsureLoginIsNotThrottled` - we instead use the limiter the user defines which is assigned to the route in Fortify's route file.

@taylorotwell - I'm sorry but I disagree. To begin all the confusion, the limiter that is set in the config is the same one used in the route in Fortify's route file. That alone is something I proposed in my PR #544 which was closed, to have two separate config parameters. Second, unless that limiter is totally empty or `false`(bool), it will never get to that throttle handler. 

See what I did in my PR, that would solve any confusion or conflicts and provide full granular control of what I want in the middleware and what I want in Fortify's throttle handler.

So the problem is, if I remove the limiter value as you are proposing, I still have a throttle in Fortify without any control over it. That is at minimum misleading IMO.

I kindly ask you to check what I did in my PR #544. I'll have it up again with the tests passing tonight.

brnquester:
@taylorotwell  - new PR opened solving all this conflicts and confusions: https://github.com/laravel/fortify/pull/547

brnquester:
@crynobone - Fix and improvement to that issue I reported: https://github.com/laravel/fortify/pull/547

PR #547, opened 2024-06-20
Fixed issue with login throttle condition and conflict with Laravel middleware

brnquester:
# Overview 

This change will fix an issue where the `EnsureLoginIsNotThrottled` verification is never reached if `fortify.limiters.login` has any value different than `false` or totally empty. The current ternary condition logic is inverted. Main issue report here [#543 ](https://github.com/laravel/fortify/issues/543).

Also, it solves the conflict of the Laravel throttle middleware cutting the flow to Fortify throttle's own handlers by adding more granular control over the rate limits for Fortify, Laravel middleware, and the decay in seconds. 

Finally, this solution will keep both throttle protections of Fortify's own handler and Laravel default middleware. If the configs proposed here are followed, the Fortify `EnsureLoginIsNotThrottled` will be triggered first returning a friendly message `Too many requests. Please try again in X seconds.` which plays nicely with InertiaJS (Other issue https://github.com/laravel/fortify/issues/312) and if the requests continue replicating a brute force behaviour, then Laravel kicks in.

In the current setup, there is a lack of clarity and differentiation between Laravel middleware throttle and Fortify's own throttle handler considering one limiter config only. Also, it has been demonstrated through multiple issues reported that there is a problem with overlapping and conflicting two approaches.

## Steps to reproduce without this fix

```
'limiters' => [
        'login' => 5,
        'two-factor' => 5,
    ],
```
```
        ->middleware(array_filter([
        ]));
```

## Relevant issues found
- https://github.com/laravel/fortify/issues/543

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -15,7 +15,11 @@ return [
     'domain' => null,
     'lowercase_usernames' => false,
     'limiters' => [
-        'login' => null,
+        'login' => 5,
+        'login-middleware' => 10, 
+        'two-factor' => 5,
+        'two-factor-middleware' => 10,
+        'decay-seconds' => 60,
     ],
     'paths' => [
         'login' => null,
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -31,14 +31,14 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
             ->name('login');
     }
 
-    $limiter = config('fortify.limiters.login');
-    $twoFactorLimiter = config('fortify.limiters.two-factor');
+    $limiterMiddleware = config('fortify.limiters.login-middleware');
+    $twoFactorLimiterMiddleware = config('fortify.limiters.two-factor-middleware');
     $verificationLimiter = config('fortify.limiters.verification', '6,1');
 
     Route::post(RoutePath::for('login', '/login'), [AuthenticatedSessionController::class, 'store'])
         ->middleware(array_filter([
             'guest:'.config('fortify.guard'),
-            $limiter ? 'throttle:'.$limiter : null,
+            $limiterMiddleware ? 'throttle:'.$limiterMiddleware : null,
         ]));
 
     Route::post(RoutePath::for('logout', '/logout'), [AuthenticatedSessionController::class, 'destroy'])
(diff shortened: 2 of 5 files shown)


taylorotwell:
I don't agree that anything needs to be changed here.

brnquester:
> I don't agree that anything needs to be changed here.

@taylorotwell - OK. I'll try my best to defend the point here with facts. The current code have one config (`fortify.limiters.login`) that provides the option to configure the throttle rate. By default, it is only used in two places, in the routes which is passed to the Laravel middleware, and to Fortify's login pipeline. It is not passed to Fortify `EnsureLoginIsNotThrottled` nor `LoginRateLimiter`. The `LoginRateLimiter` has all the rates hardcoded. So what is the point of that config then? At this moment, it is to generate confusion only.

Finally, we have one config option that is mutually exclusive. If I set the `fortify.limiters.login` it doesn't run `EnsureLoginIsNotThrottled` and run Laravel Throttle, if I set it to empty or false, it breaks the Laravel middleware and uses the Fortify's handler. 

You have to agree that at minimum it is misleading and lacks a lot of clarification. You can see that through multiple issues already reported in the past.

taylorotwell:
No, I don't agree it is confusing. You only set `foritfy.limiters.login` if you want to **totally override** the entire rate limiting strategy for logins with your own user defined rate limiter. It should be a string name, not a number.

brnquester:
> No, I don't agree it is confusing. You only set `foritfy.limiters.login` if you want to **totally override** the entire rate limiting strategy for logins with your own user defined rate limiter. It should be a string name, not a number - I think this is where you are confused.
> 
> ```
> `login` => 'the-name-of-my-rate-limiter`
> ```
> 
> https://laravel.com/docs/11.x/routing#rate-limiting

@taylorotwell - I'm not confused about that. I might not be doing a good job communicating this because I feel you're hundreds of miles away from the point of my PR. The point is when and how can Fortify's throttle (verified here: `EnsureLoginIsNotThrottled`) be used so we can leverage the friendlier error message instead of the flow cut middleware. To do that, I would need to either set `foritfy.limiters.login` to false or leave it empty, which becomes confusing and with little to know flexibility to control since your rate limits are hard coded. That's all.

Anyway. I tried. Feel free to close this PR if you still don't agree. I have solved it on my end. Regardless of the disagreement, awesome package, great community, good job with the eco-system of Laravel.

crynobone:
Upon investigating on this yes @taylorotwell is correct on the matter.

The value of `fortify.limiters.login` is either `null` of the rate limiter name such as:

```php
`login` => 'the-name-of-my-rate-limiter`
```

Which later then you can define the rate limiter in your service provider such as:

```php
RateLimiter::for('the-name-of-my-rate-limiter', function (Request $request) {
    $throttleKey = Str::transliterate(Str::lower($request->input(Fortify::username())).'|'.$request->IP());

    return Limit::perMinute(5)->by($throttleKey);
});
```

https://github.com/laravel/fortify/blob/5c2e9cdf589e439feb1ed2911d4acc7ece0ec49e/stubs/FortifyServiceProvider.php#L36-L40

brnquester:
@taylorotwell @crynobone - thanks for taking the time to review and for the input. Closing the PR.

PR #556, opened 2024-07-11
[1.x] Ability to enforce 2FA

danjohnson95:
This PR adds the ability to enforce two factor authentication (2FA).

It's quite a common feature for organisation administrators to want to enforce that their users to have 2FA enabled.

**How does it work?**

When a user logs in, if they do not have 2FA enabled and the enforce 2FA feature is enabled, a user will be redirected to a "Setup two factor authentication" page before they can proceed.

For XHR requests, the response will indicate that 2FA is required, and will contain all the required information to set up 2FA.

**How is it enabled?**

It's off by default, and it's derived from `twoFactorAuthenticationEnforced()` in the `TwoFactorAuthenticatable` trait. This means it'll be easy to conditionally activate this based on the application needs.

You'll also need to enable the "enforced" option on the features:

```php
Features::twoFactorAuthentication([
  'enforced' => true,
])
```

**How do I implement this?**

Enable the `enforced` option. And optionally define custom logic in the `twoFactorAuthenticationEnforced()` method to restrict enforcement to a subset of users.

Create a new view, `auth/two-factor-setup.blade.php`, which instructs the user that they need to enable 2FA before they can proceed.

The view should contain a form which makes a POST request to `route('two-factor.setup')`. If the Fortify "confirm" feature is enabled, this form must include a field for the user to enter their 2FA code.

**Are there any breaking changes?**

No.

**Are there tests?**

Yes! 

---

I've got a branch of Laravel Jetstream which includes the required blade files, plus support for enabling 2FA on the team level. If you're interested in this feature, I'd be happy to raise a PR in the Jetstream repo too, as well as updating documentation.

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -17,7 +17,6 @@ return [
     'limiters' => [
         'login' => null,
     ],
-    'enforce_two_factor_auth' => false,
     'paths' => [
         'login' => null,
         'logout' => null,
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -17,6 +17,7 @@ return [
     'limiters' => [
         'login' => null,
     ],
+    'enforce_two_factor_auth' => false,
     'paths' => [
         'login' => null,
         'logout' => null,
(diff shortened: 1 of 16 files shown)


taylorotwell:
I think I'll let this be handled at the app level. You could easily implement a middleware that checks for this and redirects users to the 2FA setup with a message if they don't have it enabled for their account and the team has required it.

PR #558, opened 2024-07-19
[1.x] Add TwoFactorAuthenticationFailed event

antergos98:
Hi Taylor,

Thank you for Fortify. It's our first time using it heavily, and we really enjoy the package so far.

In our project, we need to log 2FA events in a database table. Although Fortify already dispatches some events, we require one that is currently missing. This PR adds that event. I believe this is a low-risk change, and I'm happy to move the event call elsewhere if you prefer.

### DIFF
--- a/src/Events/TwoFactorAuthenticationFailed.php
+++ b/src/Events/TwoFactorAuthenticationFailed.php
@@ -2,27 +2,7 @@
 
 namespace Laravel\Fortify\Events;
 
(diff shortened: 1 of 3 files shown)


taylorotwell:
Any reason not to use the abstract `TwoFactorAuthenticationEvent` class as the base class?

antergos98:
@taylorotwell Fixed

PR #559, opened 2024-07-29
Fire ValidTwoFactorAuthenticationCodeProvided Event when 2FA session is authenticated

stefanzweifel:
In an app I'm working on we would like to store a timestamp in the session, when the user last verified their 2FA[^1]. While source-diving I didn't find an event that would be fired, when the 2FA code was verified. (There are only events for when the verification failed.)

This PR adds a new event that is being fired when a new session is being authenticated using 2FA. Adding an event listener for our usecase is now trivial.

---

I was contemplating if this event should be fired within the `TwoFactorAuthenticationProvider`, but as no other events are fired within that implementation I think the controllers is the right place for them.

[^1]: In our app, the admin panel is protected with a `RequireTwoFactorAuthentication`-middleware that periodically asks users to confirm their 2FA code. (Similar to the `\Illuminate\Auth\Middleware\RequirePassword`-middleware). 
We currently run into an _UX_-issue where folks can't access the admin dashboard immediately after login, as the same 2FA code doesn't seem to be accepted by the `TwoFactorAuthenticationProvider` within the 30s window after login.

### DIFF
--- a/src/Events/TwoFactorAuthenticationVerified.php
+++ b/src/Events/TwoFactorAuthenticationVerified.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Laravel\Fortify\Events;
+
+class TwoFactorAuthenticationVerified extends TwoFactorAuthenticationEvent
+{
+    //
+}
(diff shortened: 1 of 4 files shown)


taylorotwell:
I would call the event something like `ValidTwoFactorAuthenticationCodeProvided`

taylorotwell:
Please mark as ready for review when the requested changes have been made.

stefanzweifel:
@taylorotwell Definitely sounds and feels better than my original name. Thanks!

Issue #567, opened 2024-08-30
Adding context length configuration for 2FA to ensure better security standards


MattLoyeD:
### Summary
We need to ensure a minimum length for 2FA secret, current secret length by default is 80-bit (16 characters), but 128-bit (26 characters) is becoming minimum in some cases and the best default is 160-bit.

It's recommended to use 128-bit or 160-bit because some Authenticator apps may have problems with non-RFC-recommended lengths (Namely https://play.google.com/store/apps/details?id=org.fedorahosted.freeotp).

### Proposal
Just add some contextual config('fortify-options.two-factor-authentication.secret-length', 16), it will be retro compatible and secured as well. 
In https://github.com/laravel/fortify/blob/dd2c276e3df1ac3f47a5ab248178ea35dce1b099/src/TwoFactorAuthenticationProvider.php#L43
```php
<?php
    /**
     * Generate a new secret key.
     *
     * @param  int  $secret_length = 16
     * @return string
     */
    public function generateSecretKey(int  $secret_length = 16 )
    {
        return $this->engine->generateSecretKey($secret_length);
    }
```

In https://github.com/laravel/fortify/blob/dd2c276e3df1ac3f47a5ab248178ea35dce1b099/src/Actions/EnableTwoFactorAuthentication.php#L37

```php
<?php
   /**
     * Enable two factor authentication for the user.
     *
     * @param  mixed  $user
     * @param  bool  $force
     * @return void
     */
    public function __invoke($user, $force = false)
    {
        if (empty($user->two_factor_secret) || $force === true) {
            $secret_length = (int) config('fortify-options.two-factor-authentication.secret-length', 16);
            $user->forceFill([
                'two_factor_secret' => encrypt($this->provider->generateSecretKey($secret_length)),
                'two_factor_recovery_codes' => encrypt(json_encode(Collection::times(8, function () {
                    return RecoveryCode::generate();
                })->all())),
            ])->save();

            TwoFactorAuthenticationEnabled::dispatch($user);
        }
    }
```

There is also some adaptation to do on https://github.com/laravel/fortify/blob/dd2c276e3df1ac3f47a5ab248178ea35dce1b099/src/Contracts/TwoFactorAuthenticationProvider.php#L12

driesvints:
Heya, thanks for submitting this.

[...]

PR #568, opened 2024-09-09
Adding context length configuration for 2FA to ensure better security standards

MattLoyeD:
Follow up of [https://github.com/laravel/fortify/issues/567](https://github.com/laravel/fortify/issues/567)

### Description
We need to ensure a minimum length for 2FA secret, current secret length by default is 80-bit (16 characters), but 128-bit (26 characters) is becoming minimum in some cases and the best default is 160-bit.

It's recommended to use 128-bit or 160-bit because some Authenticator apps may have problems with non-RFC-recommended lengths (Namely https://play.google.com/store/apps/details?id=org.fedorahosted.freeotp).

### Implementation
Added some contextual config via : 
config('fortify-options.two-factor-authentication.secret-length', 16);

Retro compatible and define a security standard as expected by new RFC.

### DIFF
--- a/src/Actions/EnableTwoFactorAuthentication.php
+++ b/src/Actions/EnableTwoFactorAuthentication.php
@@ -37,9 +37,11 @@ class EnableTwoFactorAuthentication
     public function __invoke($user, $force = false)
     {
         if (empty($user->two_factor_secret) || $force === true) {
-            $secret_length = (int) config('fortify-options.two-factor-authentication.secret-length', 16);
+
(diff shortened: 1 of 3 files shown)


taylorotwell:
We don't use snake case variables. We also can't change the contract on a patch release since it would break existing applications. Is there a way to just make this opinionated without breaking anything? Please mark as ready for review when the requested changes have been made.

MattLoyeD:
Thank you @taylorotwell for your precise review, I made the necessary change. We can probably even make less breaking change by just adding the config(...) in src/TwoFactorAuthenticationProvider.php directly.

PR #570, opened 2024-10-17
Fix unnamed routes when views are disabled

cima-alfa:
This PR fixes additional unnamed routes, such as login and register, when views are disabled.
It also fixes the issue with breadcrumbs potentially not working when GET routes don't have a name. (https://github.com/laravel/fortify/pull/368#issuecomment-1248084298)

In the current state, routes like login and register have to be hardcoded since they don't have names when views are disabled.


taylorotwell:
Can you resubmit without all of the code formatting changes?

cima-alfa:
Done: https://github.com/laravel/fortify/pull/571#issue-2594736901

PR #571, opened 2024-10-17
Fix unnamed routes when views are disabled (with original code formatting)

cima-alfa:
This PR fixes additional unnamed routes, such as login and register, when views are disabled.
It also fixes the issue with breadcrumbs potentially not working when GET routes don't have a name. (https://github.com/laravel/fortify/pull/368#issuecomment-1248084298)

In the current state, routes like login and register have to be hardcoded since they don't have names when views are disabled.

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -138,7 +138,7 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
         RoutePath::for('password.confirm', '/user/confirm-password'),
         [ConfirmablePasswordController::class, 'store']
     )
-    ->middleware([config('fortify.auth_middleware', 'auth').':'.config('fortify.guard')]);
+        ->middleware([config('fortify.auth_middleware', 'auth').':'.config('fortify.guard')]);
 
     if (! $enableViews) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm finding this PR a bit confusing. Why are we only naming routes when views are disabled? 😅

cima-alfa:
I am using Ziggy JS to use Laravel’s named routes in my NextJS frontend. This allows me not having to hardcode the routes in the Frontend part. Login, register and some other routes don’t have names (others do) so I am forced to hardcode only those few routes.

enrymari:
Before this update i used to modifify fortify routes.php to register login route with get method, even if views are disabled, to allow the login page (in vue).
This patch don't resolve the problem becouse assign the post method with the name 'login'.
There're some different workaround?

cima-alfa:
I don't think your case is an issue with my proposal. However I have the same use case as you in my current project.
I solved it by simply naming the get route `front.login`. So I have a POST `login` route that is part of fortify and a GET `front.login` route that I use in my Next.js frontend. I also prefixed all fortify routes with `/api` because I consider those part of my api backend.

PR #575, opened 2024-11-01
[1.x] Simplifying UpdateUserProfileInformation.php

MrPunyapal:
This pull request includes changes to the `update` method in `UpdateUserProfileInformation.php` to streamline the process of updating user profile information, particularly focusing on email verification.

### Improvements to email verification logic:

* Simplified the email verification logic by removing the `updateVerifiedUser` method and handling email verification directly within the `update` method.
* Updated the condition to check if the email has changed using the `isDirty` method and set `email_verified_at` to null if the email is modified and the user must verify their email.
* Ensured that the email verification notification is sent if the email was changed by using the `wasChanged` method.


taylorotwell:

PR #578, opened 2024-11-20
[1.x] Mark `password` args as sensitive in ConfirmPassword

cosmastech:
To avoid revealing any sensitive details in a stack trace.

### DIFF
--- a/src/Actions/ConfirmPassword.php
+++ b/src/Actions/ConfirmPassword.php
@@ -15,7 +15,7 @@ class ConfirmPassword
      * @param  string|null  $password
      * @return bool
      */
-    public function __invoke(StatefulGuard $guard, $user, ?string $password = null)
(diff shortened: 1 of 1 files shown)


src/Actions/ConfirmPassword.php
```diff
@@ -32,7 +32,7 @@ public function __invoke(StatefulGuard $guard, $user, ?string $password = null)
      * @param  string|null  $password
      * @return bool
      */
-    protected function confirmPasswordUsingCustomCallback($user, ?string $password = null)
+    protected function confirmPasswordUsingCustomCallback($user, #[\SensitiveParameter] ?string $password = null)
```
crynobone, line 35: This version requires PHP 8.1 as minimum and `SensitiveParameter` is available on PHP 8.2 and above.

Need to require https://github.com/symfony/polyfill-php82 if we want to merge this

cosmastech: Ah thanks! I'll close. 🙇‍♂️

PR #580, opened 2024-11-27
Replace implicitly nullable parameters for PHP 8.4

JeppeKnockaert:
This PR removes implicitly nullable parameters because these result in deprecation warnings for PHP 8.4.

https://www.php.net/manual/en/migration84.deprecated.php#migration84.deprecated.core.implicitly-nullable-parameter


txdFabio:
I see this PR is merged, but I don't see the changes into the current latest release, am I missing something or it's not yet released?

PR #393, opened 2022-06-25
[1.x] fix: 🐛 Retrieve through UserProvider as possible

mpyw:
## Overview

The related changes have been previously reverted, but we see no problem in merging these changes now.

- https://github.com/laravel/fortify/pull/189 by @driesvints (merged and reverted by @taylorotwell)
- https://github.com/laravel/fortify/pull/318 by @MaksTech (closed by @driesvints)

## Description

- The original PR #189 by @driesvints might have some bugs. I mentioned [here](https://github.com/laravel/fortify/pull/189#discussion_r906661504).
  > The verification might be not enough. `retrieveByCredentials()` does not expect password matches, only an email address does. Password hash validation is done by `validateCredentials()`, but its call was not included in the PR.
  > 
  > I think it is the cause of the PR reverted. 
  > 
  > @taylorotwell Is it right?
  
  Although no reason was given in the PR and commits, it might have lead to the revert.
- Re-created PR #318 by @MaksTech is looks good to me, but @driesvints appears to have closed it for unknown reasons.
- We've got stuck at https://github.com/mpyw/scoped-auth/issues/23 because of this, so I decided to re-re-create this PR with detailed explanations.

PR #582, opened 2024-12-19
[1.x] Use `retrieveByCredentials()` on the User Provider instead of a hardcoded Eloquent query

pascalbaljet:
The `AttemptToAuthenticate` action calls the `attempt()` method on the guard, which calls `retrieveByCredentials()` under the hood. Unfortunately, the `RedirectIfTwoFactorAuthenticatable` does not use this method but instead uses a hardcoded query on the Eloquent Model, effectively skipping any logic that might be implemented in a custom User Provider.

This PR changes the `RedirectIfTwoFactorAuthenticatable` action by also using the `retrieveByCredentials()` method, making it consistent with the `AttemptToAuthenticate` action and allowing for custom User Providers to be used.

In our case, we don't have such a complex custom User Provider; we just extend the default `EloquentUserProvider` and override the `retrieveByCredentials()` method to query the username against two columns in the database (email + username).

### DIFF
--- a/src/Actions/RedirectIfTwoFactorAuthenticatable.php
+++ b/src/Actions/RedirectIfTwoFactorAuthenticatable.php
@@ -86,17 +86,17 @@ class RedirectIfTwoFactorAuthenticatable
             });
         }
 
-        $model = $this->guard->getProvider()->getModel();
+        $provider = $this->guard->getProvider();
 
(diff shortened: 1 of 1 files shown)


mpyw:
@pascalbaljet Thank you for re-submitting #393 😆

PR #585, opened 2025-01-07
Add schema check for two factor authentication columns in migration file

alibori:
Hi 👋 

I propose this change in the migration file to avoid "existing columns" errors when installing `laravel/fortify` in an existing project.

Thank you very much.


taylorotwell:

PR #593, opened 2025-02-18
Add rate limiter for registration route

TobMoeller:
This PR adds an (optional) rate limiter to the registration route to help mitigate abuse. Without throttling, bots or malicious users could potentially create millions of accounts in a short period or perform a denial-of-service attack by overwhelming the endpoint.

## Testing Considerations
I haven't included automated tests for this functionality yet. If someone can offer guidance or assistance on the best approach to test this rate limiter, that would be greatly appreciated.

## Documentation
The UPGRADE.md file currently includes changes to the two-factor rate limiting. I was uncertain, if this addition should be included aswell.

## Rate Limiting Logic
The current logic (5 registrations per minute per IP) is debatable. I welcome discussion on whether this limit should be adjusted, made configurable, or if alternative approaches should be considered.

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -32,6 +32,7 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
     }
 
     $limiter = config('fortify.limiters.login');
+    $registrationLimiter = config('fortify.limiters.registration');
     $twoFactorLimiter = config('fortify.limiters.two-factor');
     $verificationLimiter = config('fortify.limiters.verification', '6,1');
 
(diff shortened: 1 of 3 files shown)


taylorotwell:
This will block legitimate registrations due to IP addresses often being shared.

carlosmintfan:
@taylorotwell Isn’t the same true and even more for logins, then?

carlosmintfan:
Ah, sorry, I didn‘t think about the fact that login is throttled by combination of IP and username.

PR #594, opened 2025-02-19
Update error response of password reset, so that they reflect the correct input name for email

SanderSander:
Currently the default `FailedPasswordResetLinkRequestResponse` and `FailedPasswordResetResponse` return the error message on the hardcoded name "email".

In the Controllers `Fortify::email()` is used to validate the incoming requests, but this configuration option is not reflected in the response classes.


taylorotwell:
Breaking change.

PR #602, opened 2025-08-14
fix: add missing route name to `/user/two-factor-recovery-codes`

Barbapapazes:
Hello 👋,

The route `/user/two-factor-recovery-codes` does not have a name. That's not a real issue but with Wayfinder, there is no route generate which makes it impossible to use with `form.submit()`. The workaround is to use the `actions` instead which breaks the coherence within my project.


taylorotwell:
There is a typo in the name.

Barbapapazes:
> There is a typo in the name.

fixed 👍

PR #610, opened 2025-09-29
Regenerate session on register

valorin:
Regenerates session and CSRF token on user registration.


taylorotwell:
One thing to consider is Fortify is headless and may be used by something that doesn't use Laravel sessions normally? Or from an API that doesn't have sessions?

valorin:
Ah good point, I've copied the logic from `PrepareAuthenticatedSession`, which checks if the session exists first.

PR #612, opened 2025-10-03
Breaking Change: Revert "Validate password is a string when confirming password"

themsaid:
Because the type passed changed from `Illuminate\Http\Request` to `Laravel\Fortify\Http\Requests\ConfirmPasswordRequest`, Nova's `ConfirmablePasswordController` controller breaks because it passes a `Illuminate\Http\Request` to Fortify's `ConfirmablePasswordController`.

```
TypeError: Laravel\Fortify\Http\Controllers\ConfirmablePasswordController::store(): Argument #1 ($request) must be of type Laravel\Fortify\Http\Requests\ConfirmPasswordRequest, Illuminate\Http\Request given
```


devfrey:
@themsaid Why revert the bugfix without providing an alternative that doesn't break Nova? We could validate using `$request->validate()` instead.

PR #615, opened 2025-10-25
Extract credential validation logic and use StatefulGuard::once

jlswanson28694:
#### What this PR addresses
I am currently working on a project that requires users with passwords older than 90 days to reset them when they log in. My plan was to essentially copy `RedirectIfTwoFactorAuthenticatable` and make a few adjustments, since there are some similarities in the desired flow. After looking at `RedirectIfTwoFactorAuthenticatable::validateCredentials`, I noticed some issues in how it handles credential validation. Because it skips the guard entirely and validates directly with the provider, the `Attempting` and `Validated` events are not fired, and `rehashPasswordIfRequired` has to be manually called. Also, it appears that nowhere during the 2FA process does the `Authenticated` event fire.

To avoid having to upkeep `RedirectIfTwoFactorAuthenticatable` to match the login logic from `SessionGuard`, I thought it would be best to utilize the existing authentication logic by using `StatefulGuard::once` to validate credentials. The credential validation logic is now extracted to a new trait `ValidatesCredentials`, which can be added to any pipeline action that needs to do so (great for my expired password use-case).

#### Possible issues
I tried to implement this as to avoid breaking changes, but there are still a few possible issues.
- Some Laravel projects may be functioning on the assumption that the `Attempting`, `Validated`, and `Authenticated` events will **not** fire in `RedirectIfTwoFactorAuthenticatable`.
- `AttemptToAuthenticate` used to use `StatefulGuard::attempt`, which would have fired all those events a second time. I switched it to use `StatefulGuard::login` to bypass those events. However, some Laravel projects may have their own implementation of `AttemptToAuthenticate` that will still fire the events a second time.
- When `StatefulGuard::once` is triggered in `RedirectIfTwoFactorAuthenticatable`, it will authenticate the user for the remainder of the current request only. Some Laravel projects may function under the assumption that there will be no authenticated user until `AttemptToAuthenticate`. An alternative might be to immediately clear the validated user after `StatefulGuard::once`, and pass the validated user through the authentication pipeline using a static property on the `Fortify` class.
- The new `ValidatesCredentials` trait assumes that the `$guard` and `$limiter` properties exist on the parent class. I can't think of any alternatives that wouldn't introduce a possible breaking change. Maybe class inheritance instead of a trait would work, but it didn't seem appropriate for this.

### DIFF
--- a/src/Actions/AttemptToAuthenticate.php
+++ b/src/Actions/AttemptToAuthenticate.php
@@ -2,14 +2,14 @@
 
 namespace Laravel\Fortify\Actions;
 
-use Illuminate\Auth\Events\Failed;
-use Illuminate\Contracts\Auth\StatefulGuard;
-use Illuminate\Validation\ValidationException;
-use Laravel\Fortify\Fortify;
 use Laravel\Fortify\LoginRateLimiter;
+use Laravel\Fortify\ValidatesCredentials;
+use Illuminate\Contracts\Auth\StatefulGuard;
 
 class AttemptToAuthenticate
 {
+    use ValidatesCredentials;
+
     /**
      * The guard implementation.
      *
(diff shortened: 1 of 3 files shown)


taylorotwell:
I think this is probably too risky for us to merge on a patch release unfortunately.

jlswanson28694:
Makes sense, I'll work on opening a PR with the master branch.

PR #618, opened 2025-11-14
Extract credential validation logic and use StatefulGuard::once

jlswanson28694:
This is an adjusted version of PR #615 , intended for the master branch. Since breaking changes were unavoidable, I took a slightly different approach and extracted the credential validation logic to a new class `CredentialsValidator` so I could be explicit about the `StatefulGuard` and `LoginRateLimiter` requirements.

#### Purpose of this PR
1. Extract the credential validation logic to a reusable class `CredentialsValidator` that can be utilized on any step in the authentication pipeline. My use-case in (#615) was a step that redirects users with expired passwords to create a new one before logging in.
2. Within this new class, actually use the `StatefulGuard` implementation to validate credentials instead of duplicating the logic from `SessionGuard`, as `RedirectIfTwoFactorAuthenticatable` was basically doing before. I used `StatefulGuard::once` so the user is only authenticated for the remainder of the request.

#### Possible breaking changes
- `RedirectIfTwoFactorAuthenticatable` and `AttemptToAuthenticate` are significantly leaner in this PR. For anyone who is extending these two classes, the methods they are overriding may no longer exist.
- After  `CredentialsValidator::validateCredentials` is executed for the first time in the authentication pipeline, the validated user will be authenticated for the remainder of the request, but not fully logged in. Some projects may have a custom pipeline that assumes no authenticated user until `AttemptToAuthenticate`.
- `StatefulGuard::attempt` is no longer called as part of the authentication pipeline, only `StatefulGuard::once`. While `StatefulGuard::once` is required by the interface, some projects may not expect the switch.
- When using `SessionGuard`, the `Authenticated` event will now trigger twice. Once in `SessionGuard::once`, and once in `SessionGuard::login`.

### DIFF
--- a/src/Actions/AttemptToAuthenticate.php
+++ b/src/Actions/AttemptToAuthenticate.php
@@ -2,39 +2,26 @@
 
 namespace Laravel\Fortify\Actions;
 
-use Illuminate\Auth\Events\Failed;
-use Illuminate\Contracts\Auth\StatefulGuard;
-use Illuminate\Validation\ValidationException;
-use Laravel\Fortify\Fortify;
-use Laravel\Fortify\LoginRateLimiter;
+use Laravel\Fortify\CredentialsValidator;
 
 class AttemptToAuthenticate
 {
     /**
-     * The guard implementation.
+     * The credentials validator instance.
      *
-     * @var \Illuminate\Contracts\Auth\StatefulGuard
+     * @var \Laravel\Fortify\CredentialsValidator
      */
-    protected $guard;
-
-    /**
-     * The login rate limiter instance.
-     *
-     * @var \Laravel\Fortify\LoginRateLimiter
-     */
-    protected $limiter;
+    protected $validator;
 
     /**
      * Create a new controller instance.
      *
-     * @param  \Illuminate\Contracts\Auth\StatefulGuard  $guard
-     * @param  \Laravel\Fortify\LoginRateLimiter  $limiter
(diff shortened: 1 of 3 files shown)


taylorotwell:
I really am not sure these changes are necessary to Fortify internally just to support your original use case:

> I am currently working on a project that requires users with passwords older than 90 days to reset them when they log in. 

Can you not simply have a middleware that checks all authenticated requests against a `password_last_updated_at` column and then redirect them through some password update flow from there if it's > 90 days?

jlswanson28694:
To clarify, I am aware my changes are not a requirement for the original use case; I was trying to address a larger issue.

My plan was to make the expired password flow match the existing Fortify 2FA flow: when a user attempts to login, validate the user's credentials, then redirect to change their password if successful. It's already possible to add a custom action to the authentication pipeline, which is great, but I noticed a few things while I was working on this:
1. `RedirectIfTwoFactorAuthenticatable` and `AttemptToAuthenticate` share a lot of identical logic, and I was about to add yet another action that does basically the same thing. That struck me as an indication to extract the logic instead of just repeating it again.
2. If a user has 2FA enabled, the `Attempting` and `Validated` events aren't triggered before the redirect.
3. If any projects have a custom `StatefulGuard` implementation, `RedirectIfTwoFactorAuthenticatable` won't use it; `StatefulGuard` is bypassed entirely when validating credentials. Such projects would be required to use a custom authentication callback to use their `StatefulGuard` implementation during the 2FA check.

PR #624, opened 2025-11-23
Allow customizing all responses

### LINKED ISSUE #623: Not all controllers support overriding the response
### Fortify Version

1.32.0

### Laravel Version

12.39.0

### PHP Version

8.4.14

### Database Driver & Version

_No response_

### Description

There are a handful of controllers that do not support overriding the response because the response is hard-coded in the controller.

- `ConfirmedPasswordStatusController::show`
- `EmailVerificationNotificationController::show` in the `$request->user()->hasVerifiedEmail()` case
- `TwoFactorQrCodeController::show`
- `TwoFactorSecretKeyController::show`
- `RecoveryCodeController::index`

Especially the TwoFactorSecretKeyController is a bit annoying as the response uses a camelCase property which does not match the snake_case of everything else in the API.

I am overriding the routes to make the endpoints more "RESTful" and resource-oriented and would like all successful responses to be wrapped in `data`.

### Steps To Reproduce

Nothing to reproduce, this is an architectural issue.

smares:
Allows overriding all responses and closes https://github.com/laravel/fortify/issues/623

The existing response behavior was moved to the response classes introduced with this merge request.

### DIFF
--- a/src/Contracts/ConfirmedPasswordStatusResponse.php
+++ b/src/Contracts/ConfirmedPasswordStatusResponse.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Laravel\Fortify\Contracts;
+
+use Illuminate\Contracts\Support\Responsable;
+
+interface ConfirmedPasswordStatusResponse extends Responsable
+{
+    //
+}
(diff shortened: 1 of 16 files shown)


taylorotwell:

smares:
I don't know how releasing a package would help since the that was the whole purpose of this MR to make something overridable that is currently hard-coded in Fortify. With this you're forcing users to duplicate controller code just to be able to override some of the responses. And it makes no sense why some responses are overridable and some are not.

Issue #625, opened 2025-11-27
Two Additional Routes


rcravens:
### Fortify Version

latest

### Laravel Version

latest

### PHP Version

8

### Description

I recently combined Laravel Fortify with Laravel Breeze front-end components. After completing that integration, noticed that there were two routes where I needed to have the ProfileController from Laravel Breeze to manage those routes. It would be nice to add those into Laravel Fortify. Here are the routes:

GET /profile  (profile.edit) - to show the profile page for the user
DELETE /profile (profile.destroy) - to delete the user and their profile

Adding these, I believe, "completes" the route list for Fortify.

Sorry that I am not able to contribute these changes myself.

### Steps To Reproduce

NA

rcravens:
I created a pull request to add these routes.

PR #626, opened 2025-11-28
#625 Add Two Additional Routes

rcravens:
I recently combined Laravel Fortify with Laravel Breeze front-end components. After completing that integration, noticed that there were two routes where I needed to have the ProfileController from Laravel Breeze to manage those routes. It would be nice to add those into Laravel Fortify. Here are the routes:

GET /profile (profile.edit) - to show the profile page for the user
DELETE /profile (profile.destroy) - to delete the user and their profile

Adding these, I believe, "completes" the route list for Fortify.

Adding these changes.

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -106,9 +106,9 @@ Route::group( [ 'middleware' => config( 'fortify.middleware', [ 'web' ] ) ], fun
     // Profile Information...
     if ( Features::enabled( Features::updateProfileInformation() ) )
     {
-        Route::get( RoutePath::for( 'user-profile', '/user/profile' ), [ ProfileInformationController::class, 'create' ] )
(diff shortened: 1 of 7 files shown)


rcravens:
If this pull gets accepted, then I can modify my Laravel Breeze pull (https://github.com/laravel/breeze/pull/476) to get rid of the ProfileController class and fully utilize Laravel Fortify routes.

smares:
Why do you modify existing source files with just code style changes? Stick to the code style of the project.

rcravens:
> Why do you modify existing source files with just code style changes? Stick to the code style of the project.

My bad. I thought I would knock out this change quickly and pulled it into an environment where I have my own coding styles enabled.

taylorotwell:

rcravens:
I understand.

I won't release this as a package since Laravel Breeze is maintained by the Laravel Team and I would be actively merging in those changes to keep my packaged synced. I don't have the time to take on additional projects right now.

PR #627, opened 2025-12-06
Add name_prefix configuration option to Fortify

Sagmedjo:
This PR adds support for a configurable `name_prefix` for all Fortify routes.

### Why

* Enables clean usage of Fortify inside **auth route groups**.
* Prevents **route name collisions**.
* Produces **clean, separated imports with Laravel Wayfinder**.

### Wayfinder Example

**Config:**

```php
'name_prefix' => 'auth',
```

**Before:**

```ts
import { store } from '@/routes/login';
import { request } from '@/routes/password';
```

**After:**

```ts
import { store } from '@/routes/auth/login';
import { request } from '@/routes/auth/password';
```

### Implementation

* Adds `name_prefix` to the Fortify config and stub.
* Applies it via the `as` option on the Fortify route group.
* Normalizes formatting using `Str`.

### Compatibility

* Fully opt-in.
* No breaking changes.

### DIFF
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -12,6 +12,7 @@ return [
     'views' => true,
     'home' => '/home',
     'prefix' => '',
+    'name_prefix' => '',
     'domain' => null,
     'lowercase_usernames' => false,
     'limiters' => [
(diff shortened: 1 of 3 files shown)


taylorotwell:

Sagmedjo:
Hi, would be just good to know if this Feature is not wanted at all or if the PR was just not good enough.

Just afterwards I have seen, that tests in the starter kits would also have to be updated

PR #632, opened 2026-02-09
Allow UpdatesUserProfileInformation to receive Request instance instead of array

rdehnhardt:
This PR adds support for receiving an `Illuminate\Http\Request` instance in the `UpdatesUserProfileInformation` implementation, while maintaining full backward compatibility with the existing `array` signature.

## Problem

The current `ProfileInformationController` always passes `$request->all()` to the updater, which strips away access to the original Request object. This makes it impossible to use `$request->file()`, `$request->headers`, `$request->ip()`, or any other Request-specific functionality inside the 
updater implementation.

Developers who need file uploads (e.g., avatar), request metadata, or custom validation via Form Requests are forced to work around this limitation by accessing the request through the service container or other indirect means.

## Solution

The controller now inspects the second parameter type of the `update()` method via Reflection. If the implementation declares `Illuminate\Http\Request` (or a subclass), the full Request object is passed. Otherwise, `$request->all()` is passed as before.

```php
// Before (only option):
public function update(User $user, array $input): void

// After (also supported):
public function update(User $user, Request $input): void
```

## Backward Compatibility

- Existing implementations using `array $input` continue to work unchanged.
- Mocks and dynamic proxies gracefully fall back to `array` via try-catch.
- No changes to the `UpdatesUserProfileInformation` contract interface.
- Zero breaking changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
+.idea
--- a/src/Http/Controllers/ProfileInformationController.php
+++ b/src/Http/Controllers/ProfileInformationController.php
@@ -8,18 +8,19 @@ use Illuminate\Support\Str;
 use Laravel\Fortify\Contracts\ProfileInformationUpdatedResponse;
 use Laravel\Fortify\Contracts\UpdatesUserProfileInformation;
 use Laravel\Fortify\Fortify;
+use ReflectionException;
+use ReflectionMethod;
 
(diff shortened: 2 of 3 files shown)


taylorotwell:
Just use `request()->file`

rdehnhardt:
> Just use `request()->file`

I did exactly that, but with that workaround I lose dependency injection, since I use the user update service globally — not only for Fortify. But I’ve already created a new route for this. Thanks for your attention.

PR #640, opened 2026-02-16
Clarify Two-Factor Authentication JSON responses for SPA mode

mohammadRezaei1380:
This PR clarifies the JSON response behavior of two-factor authentication when Fortify is used in SPA mode (`'views' => false`).

When views are disabled, Fortify returns JSON responses instead of redirects. If a user with two-factor authentication enabled attempts to log in, the login request will indicate that a two-factor challenge is required.

This update adds documentation explaining:

- The JSON response when 2FA is required
- How the client should handle the `/two-factor-challenge` endpoint
- The expected flow for SPA authentication

This helps frontend developers better understand the authentication flow when building SPAs.

### DIFF
--- a/resources/boost/skills/developing-with-fortify/SKILL.md
+++ b/resources/boost/skills/developing-with-fortify/SKILL.md
@@ -83,6 +83,17 @@ Enable in `config/fortify.php` features array:
 
 > Use `search-docs` for integration and SPA authentication patterns.
 
+#### Two-Factor Authentication in SPA Mode
+
+When `views` is set to `false`, Fortify returns JSON responses instead of redirects.
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you share the before / after agent behavior you experienced here?

mohammadRezaei1380:
Thanks for reviewing!  

Before this PR, when Fortify was used in SPA mode (`'views' => false`), the 2FA flow used redirects internally, which didn’t translate cleanly to JSON responses — making it unclear for frontend clients how to proceed.  

With this update, the expected JSON responses and flow for handling the `/two-factor-challenge` endpoint are clearly documented, showing:
- What JSON is returned when 2FA is required
- How frontend should detect and handle it
- The proper authentication flow for SPA clients

I can add actual example responses if that would help clarify the behavior further.

PR #654, opened 2026-02-24
Update docblock return type for email verification store method

mohammadRezaei1380:
This PR updates the docblock return type for the `store` method in the email verification controller.

- Previous docblock: `@return \Illuminate\Http\Response`
- Updated docblock: `@return mixed`

The change reflects the actual return value of the method, which can be either a JSON response or a custom redirect object, depending on the request type.

This is a docs-only, low-risk update that improves type clarity and static analysis accuracy.


taylorotwell:
There are too many of these PRs being sent one at a time. If you are working on Fortify issues, just send one PR.

PR #657, opened 2026-02-26
Rename Skill

pushpak1300:


JoshSalway:
What benefits does this have?

taylorotwell:
@JoshSalway we're making sure all of our skills follow the same naming convention. @pushpak1300 works at Laravel.

PR #658, opened 2026-03-06
Add two-factor code verification as alternative to password confirmation

perepaul:
## Summary

This adds a complete two-factor code verification flow that serves as a drop-in alternative to password confirmation. Instead of requiring a password to access sensitive actions, users can verify their identity by providing a TOTP code from their authenticator app.

**New action:**
- `VerifyTwoFactorCode` — returns `bool` after verifying a TOTP code against the user's two-factor secret. Respects the confirmation flow by checking `hasEnabledTwoFactorAuthentication()`.

**New controllers (following existing naming conventions):**
- `VerifiableTwoFactorCodeController` — shows the verification view and handles code submission, storing a session timestamp on success (mirrors `ConfirmablePasswordController`).
- `VerifiedTwoFactorCodeStatusController` — returns JSON indicating whether the user has recently verified their two-factor code (mirrors `ConfirmedPasswordStatusController`).

**New contracts and responses:**
- `TwoFactorCodeVerifiedResponse` / `FailedTwoFactorCodeVerificationResponse`
- `VerifyTwoFactorCodeViewResponse` (integrated into `SimpleViewResponse`)

**Routes added (auth middleware only — no `password.confirm`, since this replaces it):**
- `GET /user/verify-two-factor-code` — view
- `POST /user/verify-two-factor-code` — store
- `GET /user/verified-two-factor-code-status` — status

**Refactor:**
- `TwoFactorLoginRequest::hasValidCode()` now delegates to the `VerifyTwoFactorCode` action, centralizing TOTP verification logic.

## Benefit

Users with two-factor authentication enabled can use their TOTP code as an alternative "sudo" check for sensitive operations, providing a consistent security UX without falling back to password re-entry.

All response contracts are swappable via the container, following the established Fortify pattern.

## Test plan

- [x] Valid TOTP code returns 200
- [x] Invalid code returns 422
- [x] Missing code returns 422
- [x] Missing two-factor secret returns 422
- [x] Unauthenticated request returns 401
- [x] Confirmed 2FA flow: valid code with `two_factor_confirmed_at` returns 200
- [x] Confirmed 2FA flow: valid code without `two_factor_confirmed_at` returns 422
- [x] Session timestamp is set on successful verification
- [x] Status endpoint returns `confirmed: true` when recently verified
- [x] Status endpoint returns `confirmed: false` when not verified
- [x] All 108 existing + new tests pass

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -18,6 +18,8 @@ use Laravel\Fortify\Http\Controllers\TwoFactorAuthenticatedSessionController;
 use Laravel\Fortify\Http\Controllers\TwoFactorAuthenticationController;
 use Laravel\Fortify\Http\Controllers\TwoFactorQrCodeController;
 use Laravel\Fortify\Http\Controllers\TwoFactorSecretKeyController;
+use Laravel\Fortify\Http\Controllers\VerifiableTwoFactorCodeController;
+use Laravel\Fortify\Http\Controllers\VerifiedTwoFactorCodeStatusController;
 use Laravel\Fortify\Http\Controllers\VerifyEmailController;
 use Laravel\Fortify\RoutePath;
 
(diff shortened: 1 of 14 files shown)


taylorotwell:

PR #660, opened 2026-03-14
 feat: Add mustConfirmTwoFactorAuthentication method

SantosVilanculos:
This PR introduces the mustConfirmTwoFactorAuthentication() method to the TwoFactorAuthenticatable trait. This method determines if a user's two-factor authentication has been enabled but is awaiting a final confirmation step.

It provides a consistent point for developers to check this state, allowing for varied UI and flow implementations to guide users through 2FA confirmation. It also eliminates the need for individual developers to implement this common logic, reducing boilerplate.


taylorotwell:

PR #665, opened 2026-03-29
Make user contracts generic

jivanf:
The contracts for the actions that interact with users currently specify `\Illuminate\Foundation\Auth\User` as the user type. However, this definition doesn't allow other user models that don't extend `\Illuminate\Foundation\Auth\User`, which leads to PHPStan errors, as shown in this example:

https://phpstan.org/r/c0fd079b-3cbf-4fa6-91be-d35bfc7f791c

By adding generics to these contracts, the actions can use any user model (as long as it extends `\Illuminate\Database\Eloquent\Model`):

https://phpstan.org/r/e5721cf1-e7f2-4803-b97f-e5c2d14e728c

This is not a breaking change because `\Illuminate\Foundation\Auth\User` is used as the default type.

### DIFF
--- a/src/Contracts/CreatesNewUsers.php
+++ b/src/Contracts/CreatesNewUsers.php
@@ -2,13 +2,16 @@
 
 namespace Laravel\Fortify\Contracts;
 
+/**
+ * @template TUser of \Illuminate\Database\Eloquent\Model = \Illuminate\Foundation\Auth\User
+ */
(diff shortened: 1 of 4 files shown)


jivanf:
Thank you for taking a look at the PR :smiley:

Is there a reason why it was closed? We currently have a patch for this in our project and I believe others could benefit from this as well.

PR #666, opened 2026-04-14
Add paths configuration to published config stub

pushpak1300:
The `paths` config key exists internally in Fortify and powers route path customization via `RoutePath::for()`, but it was missing from the published config stub. Users had no way to discover or customize route paths without reading the source.

### Usage

```php
// config/fortify.php
'paths' => [
    'login' => '/sign-in',
],
```

### Approach

- Added the full `paths` array to the published config stub, matching the internal config structure
- All values default to `null` so Fortify's built-in paths are used unless explicitly overridden


taylorotwell:
Ah, there are just too many paths here imo.

PR #668, opened 2026-04-23
Feat/add passkeys

benbjurstrom:
This PR adds support for passkeys to Fortify by wrapping the new [laravel/passkeys](https://packagist.org/packages/laravel/passkeys) composer package. For the frontend it can be paired with the official [`@laravel/passkeys`](https://www.npmjs.com/package/@laravel/passkeys) JavaScript client.

Passkeys can be enabled in the features section of `config/fortify.php`

```php
'features' => [
    // ...
    Features::passkeys(),
],
```

and configured via a dedicated `passkeys` block

```php
'passkeys' => [
    'relying_party_id' => parse_url(config('app.url'), PHP_URL_HOST),
    'allowed_origins' => [config('app.url')],
    'user_handle_secret' => config('app.key'),
    'timeout' => 60000,
],
```

The User model should implement `Laravel\Fortify\Contracts\PasskeyUser` and use the `Laravel\Fortify\PasskeyAuthenticatable` trait.

```php
use Laravel\Fortify\Contracts\PasskeyUser;
use Laravel\Fortify\PasskeyAuthenticatable;

class User extends Authenticatable implements PasskeyUser
{
    use Notifiable, PasskeyAuthenticatable;
}
```

Seven new endpoints are added for login (GET `/passkeys/login/options`, POST `/passkeys/login`), password confirmation (GET `/passkeys/confirm/options`, POST `/passkeys/confirm`), and passkey management (GET `/user/passkeys/options`, POST `/user/passkeys`, DELETE `/user/passkeys/{passkey}`)

A draft update to the Fortify docs with more information can be found here: https://github.com/laravel/docs/compare/fortify-passkeys?expand=1

### DIFF
--- a/routes/routes.php
+++ b/routes/routes.php
@@ -178,46 +178,38 @@ Route::group(['middleware' => config('fortify.middleware', ['web'])], function (
 
     // Passkeys...
     if (Features::enabled(Features::passkeys())) {
-        Route::get(RoutePath::for('passkey.login-options', '/passkeys/login/options'), [PasskeyLoginController::class, 'index'])
-            ->middleware(array_filter([
-                'guest:'.config('fortify.guard'),
-                $passkeyLimiter ? 'throttle:'.$passkeyLimiter : null,
-            ]))->name('passkey.login-options');
-
-        Route::post(RoutePath::for('passkey.login', '/passkeys/login'), [PasskeyLoginController::class, 'store'])
-            ->middleware(array_filter([
-                'guest:'.config('fortify.guard'),
-                $passkeyLimiter ? 'throttle:'.$passkeyLimiter : null,
-            ]))->name('passkey.login');
-
-        $passkeyAuthMiddleware = [config('fortify.auth_middleware', 'auth').':'.config('fortify.guard')];
-
-        Route::get(RoutePath::for('passkey.confirm-options', '/passkeys/confirm/options'), [PasskeyConfirmationController::class, 'index'])
-            ->middleware(array_filter(array_merge($passkeyAuthMiddleware, [
-                $passkeyLimiter ? 'throttle:'.$passkeyLimiter : null,
-            ])))
-            ->name('passkey.confirm-options');
-
-        Route::post(RoutePath::for('passkey.confirm', '/passkeys/confirm'), [PasskeyConfirmationController::class, 'store'])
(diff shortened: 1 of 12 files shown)


src/FortifyServiceProvider.php
```diff
@@ -110,6 +113,60 @@ protected function registerResponseBindings()
...
+     */
+    protected function configurePasskeys()
+    {
+        LaravelPasskeys::ignoreRoutes();
+
+        if ($model = $this->passkeyUserModel()) {
+            LaravelPasskeys::useUserModel($model);
+        }
+
+        config([
```
joetannenbaum, line 129: Does this override the published underlying passkeys config? Is that odd?

benbjurstrom: It would override a published `config/passkeys.php`. But a normal `fortify:install` would not publish the passkeys config file.

It is a little odd if someone published that config separately, but if we don’t override it here, users would have to manually keep things like guard, middleware, redirect, and throttling in sync with `config/fortify.php`.

My intention was to keep the underlying passkeys package hidden from Fortify users and make config/fortify.php the single place to configure things. If we go that route, we could add a note to the Fortify docs explaining the override.

joetannenbaum: Yeah we should at least add a note in the docs. 👍

routes/routes.php
```diff
@@ -171,4 +175,45 @@
             ->middleware($twoFactorMiddleware)
             ->name('two-factor.regenerate-recovery-codes');
     }
+
+    // Passkeys...
+    if (Features::enabled(Features::passkeys())) {
+        $throttle = $passkeyLimiter ? ['throttle:' . $passkeyLimiter] : [];
```
joetannenbaum: Can't we use `config('passkeys.throttle')` here from the service provider? Or we should keep this separate?

benbjurstrom: We could use `config('passkeys.throttle')` since the service provider syncs it from the Fortify config. But I feel like that makes it less obvious where the value is coming from. Plus we already have an established pattern declaring the limiters as local variables.

https://github.com/laravel/fortify/blob/c2f530c47ec27136fc1d7e5ec1e6e8b270fe12e1/routes/routes.php#L37-L40

Radiergummi:
@benbjurstrom This is amazing work. Thanks a lot! ❤️

PR #694, opened 2026-08-07
Add option to override RecoveryCode class

TobiasNagel01:
For a personal project, I need my recovery codes to be only upper case letters.
I tried to find a way to manipulate them, but I think the best way would be, to be able to override the class that contains the generate method.

(This is my first ever contribution to a package, if I made any mistakes, please let me know :) )


taylorotwell:
Simplified it a bit.

PR #695, opened 2026-08-11
[1.x] Preserve pending two-factor setup on repeat enable

### LINKED ISSUE #693: Repeat "Enable 2FA" click deletes the pending two_factor_secret, leaving setup in an unrecoverable 404 state
### Fortify Version

v1.37.2 (the three files involved are byte identical on v1.37.3, the current release)

### Laravel Version

13.19.0

### PHP Version

8.4.22

### Database Driver & Version

MariaDB 11.8.8. Not database dependent.

### Description

With `Features::twoFactorAuthentication(['confirm' => true])`, clicking "Enable 2FA" a second time within the same page session **destroys** the pending, unconfirmed `two_factor_secret` instead of doing nothing. The setup modal then reopens and requests `GET /user/two-factor-secret-key`, which returns 404 because the secret it is asking for was deleted moments earlier by that same click.

The result is a dead end: no QR code, no setup key, and nothing in the UI suggesting the only recovery (a full page reload).

This is an interaction between two individually reasonable behaviours introduced 19 months apart:

1. #518 (merged 2024-01-31) made `EnableTwoFactorAuthentication` a no-op when a secret already exists and `force` is false.
2. #604 (merged 2025-08-29) added `InteractsWithTwoFactorState`, which deletes an unconfirmed secret on any request whose session `two_factor_confirming_at` is stale.

Together they produce a click that enables nothing and deletes everything.

#### Sequence

Starting from a security settings page whose form request uses `InteractsWithTwoFactorState`, which is what the official starter kits do:

1. Click "Enable 2FA". `POST /user/two-factor-authentication` writes the secret. The redirect back to the settings page runs `ensureStateIsValid()`, which sees a fresh unconfirmed secret and stamps `two_factor_confirming_at = T1`. The modal fetches the QR code and setup key, both 200. This works correctly.
2. Close the modal without confirming. In the starter kits this resets the client side setup state, so the button reverts from "Continue setup" to "Enable 2FA".
3. Click "Enable 2FA" again, at least one second later. `EnableTwoFactorAuthentication` no-ops (per #518) because the secret is still present. The redirect back to the settings page runs `ensureStateIsValid()` again, where `neverFinishedConfirmingTwoFactorAuthentication()` is now true (`two_factor_confirming_at` is `T1`, not the current `time()`), so `DisableTwoFactorAuthentication` nulls `two_factor_secret`.
4. The modal opens on success and fetches the setup data. `TwoFactorSecretKeyController::show()` aborts 404 because `two_factor_secret` is now null.

#### Two related observations

`TwoFactorQrCodeController::show()` returns `[]` with a **200** in the same situation, rather than the 404 the secret key controller returns. The QR pane therefore renders an indefinite loading state instead of an error, so the two endpoints disagree about how to report the same condition.

A full page reload does recover the flow, but it issues a **different** secret. A user who already scanned the first QR code before hitting this is left with a dead entry in their authenticator app sitting next to the working one, under the same account name.

#### Failing test

This runs against a stock React or Vue starter kit application (route name `security.edit`, whose form request uses `InteractsWithTwoFactorState`). No JavaScript is involved; the sequence is pure HTTP.

```php
public function test_a_repeat_enable_deletes_the_pending_two_factor_secret(): void
{
    Features::twoFactorAuthentication(['confirm' => true, 'confirmPassword' => true]);

    $user = User::factory()->create();

    $this->actingAs($user)->withSession(['auth.password_confirmed_at' => time()]);

    // Attempt 1: works. The redirect back stamps two_factor_confirming_at.
    $this->get(route('security.edit'))->assertOk();
    $this->post(route('two-factor.enable'));
    $this->get(route('security.edit'))->assertOk();
    $this->get(route('two-factor.secret-key'))->assertOk();

    $this->assertNotNull($user->fresh()->two_factor_secret);

    sleep(1);

    // Attempt 2: the user closed the modal and clicked "Enable 2FA" again.
    $this->post(route('two-factor.enable'));        // no-op, the secret still exists
    $this->get(route('security.edit'))->assertOk();  // this request deletes the secret

    $this->assertNotNull($user->fresh()->two_factor_secret); // FAILS: it is now null

    $this->get(route('two-factor.secret-key'))->assertOk();  // FAILS: 404
}
```

Observed when instrumented:

```
[ATTEMPT 1] secret after redirect-back GET: SET     GET two-factor.secret-key => 200
[ATTEMPT 2] secret before enable POST : SET
[ATTEMPT 2] secret after  enable POST : SET
[ATTEMPT 2] secret after  redirect GET: NULL
[ATTEMPT 2] GET two-factor.secret-key => 404
```

#### In the browser

On a stock React starter kit with `confirm => true`:

1. Go to `/settings/security` and confirm your password.
2. Click "Enable 2FA". The modal shows a QR code and a setup key. Note the key.
3. Close the modal without entering a code.
4. Click "Enable 2FA" again.

The modal reopens showing "Something went wrong. / Failed to fetch a setup key". The network panel shows:

```
POST /user/two-factor-authentication -> 200
GET  /user/two-factor-qr-code        -> 200   (empty body)
GET  /user/two-factor-secret-key     -> 404
```

`two_factor_secret` is null in the database. Reloading the page and clicking "Enable 2FA" works again, but issues a different key.

#### Expected behaviour

Clicking "Enable 2FA" while an unconfirmed secret exists should either return the existing setup data or mint a fresh one. It should not leave the user with no secret and a 404.

iamhaseebn:
Fixes #693

A repeated request to enable two-factor authentication intentionally preserves an existing unconfirmed secret, but it also leaves the previous `two_factor_confirming_at` session value in place. The next request using `InteractsWithTwoFactorState` interprets that stale value as an abandoned setup and deletes the secret.

This clears the stale confirmation marker after the enable action when confirmation is required and the user still has a pending setup. The following request can then begin confirmation again while retaining the same secret and recovery codes. Confirmed users and installations that do not require confirmation are unaffected.

The regression test covers the complete disabled-to-pending-to-repeat-enable transition and verifies that the original secret remains available.

Tests:

- Focused controller and state tests: 22 passed, 72 assertions
- Full test suite: 120 passed, 376 assertions
- PHPStan: no errors
- Composer validation: passed

### DIFF
--- a/src/Http/Controllers/TwoFactorAuthenticationController.php
+++ b/src/Http/Controllers/TwoFactorAuthenticationController.php
@@ -8,6 +8,7 @@ use Laravel\Fortify\Actions\DisableTwoFactorAuthentication;
 use Laravel\Fortify\Actions\EnableTwoFactorAuthentication;
 use Laravel\Fortify\Contracts\TwoFactorDisabledResponse;
 use Laravel\Fortify\Contracts\TwoFactorEnabledResponse;
+use Laravel\Fortify\Fortify;
 
 class TwoFactorAuthenticationController extends Controller
 {
@@ -21,6 +22,12 @@ class TwoFactorAuthenticationController extends Controller
     public function store(Request $request, EnableTwoFactorAuthentication $enable)
     {
         $enable($request->user(), $request->boolean('force', false));
+
+        if (Fortify::confirmsTwoFactorAuthentication() &&
+            ! is_null($request->user()->two_factor_secret) &&
+            is_null($request->user()->two_factor_confirmed_at)) {
+            $request->session()->remove('two_factor_confirming_at');
+        }
 
         return app(TwoFactorEnabledResponse::class);
     }
--- a/tests/TwoFactorAuthenticationControllerTest.php
+++ b/tests/TwoFactorAuthenticationControllerTest.php
@@ -5,12 +5,16 @@ namespace Laravel\Fortify\Tests;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Encryption\Encrypter;
 use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Event;
 use Illuminate\Support\Str;
+use Laravel\Fortify\Actions\EnableTwoFactorAuthentication;
 use Laravel\Fortify\Events\TwoFactorAuthenticationConfirmed;
 use Laravel\Fortify\Events\TwoFactorAuthenticationDisabled;
 use Laravel\Fortify\Events\TwoFactorAuthenticationEnabled;
+use Laravel\Fortify\Http\Controllers\TwoFactorAuthenticationController;
 use Laravel\Fortify\Tests\Models\UserWithTwoFactor;
+use Laravel\Fortify\Tests\Requests\FormRequestInteractsWithTwoFactorState;
 use Orchestra\Testbench\Attributes\DefineEnvironment;
 use Orchestra\Testbench\Attributes\ResetRefreshDatabaseState;
 use Orchestra\Testbench\Attributes\WithMigration;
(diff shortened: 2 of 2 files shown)


taylorotwell:

I appreciate you taking the time to submit this; however, it appears this contribution may have been primarily AI-generated without careful human review and consideration.

We've found that AI-generated code often doesn't align well with Laravel's conventions, architectural decisions, and the specific context of what we're trying to accomplish with the framework. Quality contributions require thoughtful human insight into the codebase.

If you're interested in contributing to Laravel, I'd encourage you to familiarize yourself with the existing codebase, engage with the community, and submit PRs that reflect your own understanding and careful consideration of the problem you're solving.

PR #697, opened 2026-09-09
Introduction of Timebox to prevent Timing-based Username Enumeration

bretto36:
Hi There, we reached out via email to Taylor using the security policy instructions and have not heard anything back. 

There is a timing based username enumeration issue with fortify, it uses the same mechanism that was discovered in laravel framework - https://nvd.nist.gov/vuln/detail/cve-2022-40482

I'm not sure this is the best solution, and i'm also not sure how to test effectively. 

But at least this is demonstrating the potential solution so someone more competent can action.


taylorotwell:
I don't think this needs injection. Just new up the Timebox in the method.

bretto36:
> I don't think this needs injection. Just new up the Timebox in the method.

I was trying to match how Session Guard did it as the duration is configurable in the sessions.php config i'm pretty sure. 

Let me switch to new

